PR #1, opened 2020-03-17
First public release

prescod:
Copy over files from internal repo to public repo. 

Repo does not yet adhere to various standards of CI, branch protection etc. That's the next step.


LICENSE
```diff
@@ -0,0 +1,27 @@
+    Copyright (c) 2015, Salesforce.org
```
davisagli:
```suggestion
    Copyright (c) 2020, Salesforce.org
```

PR #2, opened 2020-03-17
Statistical distributions in the new Plugin format.

prescod:
Distributions that meet various statistical patterns. Plugin will not load if NumPy is not available, but that's probably okay for this use-case.

Also a small improvement to plugin error handling.

### DIFF
--- a/snowfakery/standard_plugins/statistical_distributions.py
+++ b/snowfakery/standard_plugins/statistical_distributions.py
@@ -1,0 +1,29 @@
+from numpy.random import normal, lognormal, binomial, exponential, poisson, gamma
+from numpy.random import seed
+
+
+from snowfakery.plugins import SnowfakeryPlugin
+
+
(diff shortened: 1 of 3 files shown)


snowfakery/standard_plugins/statistical_distributions.py
```diff
@@ -0,0 +1,29 @@
...
+def wrap(distribution):
+    def _do_distribution(self, **params):
+        random_seed = params.pop("seed", None)
+        seed(random_seed)
+        return float(distribution(**params, size=1).astype(float)[0])
+        # except (TypeError, DataGenValueError, DataGenError):
+        #     raise DataGenError(
+        #         f"Incorrect parameters {params} for {distribution.__name__} distribution.",
+        #         None, None,
+        #     )
```
davisagli: Could you add a docstring to explain what the wrapper does, and remove commented code?

PR #6, opened 2020-04-02
Change the format for i18n to something more YAMLishous.

prescod:


tests/CharacterGenTest.yml
```diff
@@ -6,7 +6,10 @@
 - object: Druid
   count: <<num_druids>>
   fields:
-    Name: <<fake_i18n('de_DE').first_name()>>
+    Name:
+      i18n_fake:
+        locale: no_NO
+        fake: first_name
```
davisagli, line 12: In terms of usability I think something like this would feel more natural:

    Name:
      locale: no_NO
      fake: first_name

i.e. locale pushes to a stack-based context which influences construction of fakers within that context (and ideally also available to non-faker functions)

But I haven't looked closely at what it would take to implement that...

prescod: I started down this path but it got messy and I decided it was better to wait for clearer requirements from users.

First, you might want to set the locale at the top level or in the middle of a list of friends and where would you do that? I'd need some kind of new pragma/directive/setting-type thing.

Which leads to a medium-sized implementation issue because internally Snowfakery doesn't work as a top-to-bottom interpreter because there's historically been only one kind of "ordered declaration" (macros etc. are basically all globally scoped).

Ironically your comment did inspire the creation of stack-based contexts for plugins, which I'm sure will be useful for various things, but in the end its too much of hassle to use them for this particular use-case until someone has a real need for them. I'm pretty sure I originally implemented internationalization just because Beth asked "could it do that?" and I thought "sure, why not? Should be 15 minutes of work."

If the requirements become clearer -- especially around desired scoping -- then its fairly easy to add a "locale" feature in a backwards compatible way and deprecate this function.

PR #9, opened 2020-04-07
Snowfakery history.

prescod:


davisagli:
@prescod Are the docs going to end up in Sphinx, or elsewhere? If they're in sphinx and we want to include the history (we should), then it'll be easiest if the history is in reStructuredText. (I don't have a problem with using Markdown if there's a good way to make it work.)

prescod:
It will be easy to switch over when I have a unified ReadTheDocs strategy.

PR #10, opened 2020-04-08
Turn on black and pre-commit hooks

prescod:
Just code reformatting so I can turn on the pre-commit hooks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r requirements_dev.txt
 
-          # - name: Code Quality
-      #   run: black . --check
+      - name: Code Quality
+        run: black --check --diff .
 
       - name: pytest
         run: pytest --cov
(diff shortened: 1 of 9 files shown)


davisagli:
@prescod Is adding the pre-commit hooks going to happen separately?

prescod:
> @prescod Is adding the pre-commit hooks going to happen separately?

There was a pre-commit hook, it just wasn't turned on in my local instance of the new  repo.

https://github.com/SFDO-Tooling/Snowfakery/blob/master/.pre-commit-config.yaml

Users need to turn it on repo by repo right? The pre-commit docs don't talk about whether it can work effectively as a "global hook".

davisagli:
@prescod yeah it typically needs to be installed for each repo. Here are the instructions for doing so for CumulusCI (step 4): https://github.com/SFDO-Tooling/CumulusCI/blob/master/CONTRIBUTING.rst#get-started.

We should also have a step in the CI github action that runs black

prescod:
Black is configured:

https://github.com/SFDO-Tooling/Snowfakery/runs/571877925#step:5:1

CCI doesn't seem to automate Flake8 so I didn't here either.

PR #12, opened 2020-04-08
Placeholder: Feature/coverage

prescod:
Increasing test coverage and Windows support and fixing bugs found in that process.

In the file cli.py, I removed the broken --start-ids feature and use a context manager for outputstreams to close them. Let Output Streams open and close their old file handles.

In datagenerator.py I allow click's LazyFile which is not a proper subclass of Python TextIO

In OutputStreams I have a base class that manages open and closing file handles. Also cleaned up connection handling somewhat in SQLOutputStream and create everything (base, metadata, etc.) in the initializer rather than in the middle of the code.

In test files, increase coverage and remove uses of NamedTemporaryFile which is basically unusable on Windows. I created a new utility (in test/utils.py) for that.

### DIFF
--- a/snowfakery/cli.py
+++ b/snowfakery/cli.py
@@ -12,6 +12,7 @@ from snowfakery.data_gen_exceptions import DataGenError
 from snowfakery.data_generator import generate, StoppingCriteria
 import sys
 from pathlib import Path
+from contextlib import contextmanager
 
 import yaml
 import click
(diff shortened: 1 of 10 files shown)


snowfakery/cli.py
```diff
@@ -89,14 +91,6 @@ def string_int_tuple(ctx, param, value: Tuple[str, str] = None):
     "generate_cci_mapping_file",
     type=click.Path(exists=False),
 )
-@click.option(
```
prescod, line 92: This option didn't actually work. It was superseded by the continuation-file feature.

snowfakery/output_streams.py
```diff
@@ -194,51 +215,42 @@ class JSONOutputStream(OutputStream):
...
 
     @classmethod
     def from_url(cls, db_url: str, mappings: Optional[Dict] = None):
-        self = cls(create_engine(db_url), mappings)
-        return self
-
-    @classmethod
-    def from_connection(
-        cls, session, engine: Engine, base, mappings: Optional[Dict] = None
-    ):
```
davisagli, line 232: from_connection is currently used in CumulusCI; I imagine you're intending to make a new release of snowfakery and then update https://github.com/SFDO-Tooling/CumulusCI/pull/1659?

snowfakery/output_streams.py
```diff
@@ -375,16 +382,24 @@ def write_single_row(self, tablename: str, row: Dict) -> None:
...
+    def __init__(self, path, format):
+        self.path = path
         self.format = format
-        super().__init__(file)
+        super().__init__(path)
 
     def close(self) -> None:
-        self.G.draw(path=self.file, prog="dot", format=self.format)
+        if isinstance(self.path, Path):
+            self.path = str()
```
davisagli, line 401: Huh?

prescod: Removed. It was both obsolete and incomplete.

PR #22, opened 2020-04-08
Weekly dependabot updates

prescod:


davisagli:
@prescod should we make setup.py list unpinned requirements (or pinned to minimum version) so that a release of Snowfakery can be used with installations of cumulusci that have updated to newer versions?

PR #23, opened 2020-04-09
Feature/faker yaml compatibility

prescod:
Faker and YAML are not compatible in some way that mysteriously only seems to be triggered (n this code base, at least) in Python 3.8.

I haven't had a chance to debug my tox situation in Snowfakery but running manually on my computer it seems to work in all 3 relevant Python versions.

`$ .tox/py36/bin/python -m pytest && .tox/py37/bin/python -m pytest &&  .tox/py38/bin/python -m pytest && echo success`

...

`success`

Although I won't generally bump the version number in feature branches, I do need to deploy this fix to unblock the CCI PR, so I'll do a release as soon as it is merged.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,10 @@ In the beginning, programmers created the databases. Now the databases were form
 
 And so [Salesforce.org](http://salesforce.org/) said “Let there be data,” and there was Snowfakery. And it was good.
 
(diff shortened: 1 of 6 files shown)


davisagli:
@prescod if you're planning to make a release from this commit, there should be an update to the history too (even if just to say "no user-facing-changes")

PR #24, opened 2020-04-10
Changes necessary for tox to work.

prescod:


prescod:
Okay I've created a pull request for that now but I'd rather commit this one and then work on that one:

https://github.com/SFDO-Tooling/Snowfakery/pull/25

PR #25, opened 2020-04-10
Placeholder: tox in GitHub

prescod:


.github/workflows/ci.yml
```diff
@@ -3,27 +3,29 @@ name: CI
...
-          pip install -r requirements_dev.txt
-
-      - name: Code Quality
-        run: black --check --diff .
+          set -xe
+          python -VV
+          python -m site
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install --upgrade coverage[toml] virtualenv tox tox-gh-actions
+          python -m pip install -r requirements_dev.txt --upgrade
```
davisagli, line 28: I'm confused...you're installing tox but not running it.

prescod: Yeah I didn't cut and paste enough and hadn't gotten around to looking closer. It looks like everything works now.

I'm a little bit fuzzy on how Coveralls is picking which build is "canonical" for the commit (first reported? last? all? any?) but it probably doesn't matter in practice.

PR #27, opened 2020-04-14
Feature/output folder

prescod:
Add a new CLI option called output-folder which determines where the output goes. Defaults to "."

Use this to drive CSV folder selection instead of the csvfile:/// convention which was confusing and broken on Windows.


prescod:
Added one tiny change @davisagli :

https://github.com/SFDO-Tooling/Snowfakery/pull/27/commits/5e8a240544ab051096084e63163aa1c958033f5b

PR #35, opened 2020-04-22
Various cleanups and reliability improvements.

prescod:
Coerce all faker values to strings (even things like latitudes, longitudes, dates).

Ignore fields starting with "__" consistently per the documentation.

Allow None as a field value (useful sometimes).

Delete obsolete file.


snowfakery/parse_factory_yaml.py
```diff
@@ -172,11 +178,7 @@ def parse_structured_value(
 def parse_field_value(
     name: str, field, context: ParseContext, allow_structured_values=True
 ) -> Union[SimpleValue, StructuredValue, ObjectTemplate]:
-    if field is None:
-        raise DataGenSyntaxError(
-            f"Field should not be empty: {name}", **context.line_num(field)
-        )
-    if isinstance(field, (str, Number, date)):
+    if isinstance(field, (str, Number, date, type(None))):
```
davisagli, line 181: If the value is None, how does that end up getting serialized for output?

prescod: NULL in SQL
null in JSON
'' in CSV
None in DebugOutputStream

I added some tests to show that.

https://github.com/SFDO-Tooling/Snowfakery/pull/35/commits/4c11f170ebe39674ad9e2704c319b45126853cbe

PR #38, opened 2020-04-29
Add support for Salesforce record types to Snowfakery

prescod:
Renamed some internal attributes for clarity in data_generator.py and parse_factory_yaml.py

Delete some confusing commentary in data_generator_dom.py.

Add record_type support to generate_mapping_from_factory.py

Capture the table name on TableInfo objects so that when we clone them for record type objects, we still know what table they came from.

Add tests and a test recipe.

### DIFF
--- a/snowfakery/generate_mapping_from_factory.py
+++ b/snowfakery/generate_mapping_from_factory.py
@@ -1,15 +1,78 @@
 from copy import deepcopy
 from warnings import warn
 
+from snowfakery.data_generator import ExecutionSummary
+from snowfakery.data_gen_exceptions import DataGenNameError, DataGenError
+from snowfakery.parse_factory_yaml import TableInfo
+from snowfakery.data_generator_runtime import Dependency
 
-def mapping_from_factory_templates(summary):
+
+def mapping_from_factory_templates(summary: ExecutionSummary):
     """Use the outputs of the factory YAML and convert to Mapping.yml format"""
+    record_types = generate_record_type_pseudo_tables(summary)
     dependencies, reference_fields = build_dependencies(summary.intertable_dependencies)
-    table_order = sort_dependencies(dependencies, summary.tables)
-    mappings = mappings_from_sorted_tables(
-        summary.tables, table_order, reference_fields
-    )
+    tables = {**summary.tables, **record_types}
+    table_order = sort_dependencies(dependencies, tables)
+    mappings = mappings_from_sorted_tables(tables, table_order, reference_fields,)
     return mappings
+
+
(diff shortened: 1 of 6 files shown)


snowfakery/generate_mapping_from_factory.py
```diff
@@ -1,17 +1,80 @@
...
+        # generate a pretend table for purposes of dependency sorting,
+        # lookups and mapping generation
+        record_type_pseudo_table = record_types.setdefault(
+            record_type_name, TableInfo(template.tablename)
+        )
+        record_type_pseudo_table.register(template)
+        record_type_pseudo_table.record_type = record_type_name
+
+        # copy over the dependencies from the real table
+        for dependency in list(summary.intertable_dependencies):
```
davisagli, line 63: It's non-obvious that the reason for using list() here is to avoid mutating the same thing you're iterating over. Probably worth a comment.

snowfakery/generate_mapping_from_factory.py
```diff
@@ -1,17 +1,80 @@
...
+            if dependency.table_name_from == real_table_name:
+                summary.intertable_dependencies.add(
+                    Dependency(record_type_name, *dependency[1:])
+                )
+
+        # the record type field isn't halpful for the TableInfo of the real table anymore
```
davisagli, line 69:
```suggestion
        # the record type field isn't helpful for the TableInfo of the real table anymore
```

snowfakery/generate_mapping_from_factory.py
prescod: If you remove the copy then the error message is very clear. It won't introduce a subtle bug. But I'll change the list(s) to s.copy(), which may avoid implying that I'm trying to linearize the set. 29cab7c72f3f12151f32419d4230e3eee2afcce9

prescod:
> As I believe you noted, this seems like something that should ideally be handled by the load step in cumulusci without needing to manually set up a lookup in the mapping -- since it's such a Salesforce-specific feature, it feels like handling it belongs there more than in Snowfakery. But I take it that this is a workaround for the meantime, which I'm okay with.

Right.

> It needs docs though.

Okay, the docs are updated:

"To specify a record type for a record, just put the Record Type’s API Name in a field named RecordType."

https://salesforce.quip.com/AXY5A4WGTbNP#PJUACAQ5tGE

snowfakery/generate_mapping_from_factory.py
davisagli: Fair enough, I forgot that modern Python versions give a clear error about this.

PR #47, opened 2020-05-09
Example of child objects summing to parent

prescod:
A common question is how to make child objects that sum up to a parent object. This example shows how to use Plugin features to do that and generally helps to explain the concept of stateful plugin objects.


snowfakery/plugins.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, result: Mapping):
         self.result = result
 
     def __getattr__(self, name):
-        return self.result[name]
+        return self.__dict__["result"][name]
```
prescod, line 139: Avoids the possibility of recursive getattr if result is not defined in a subclass.

Br4nd0R (approved):

PR #49, opened 2020-05-15
Sort set for deterministic behaviour.

prescod:
Deterministic sort order in mapping file output.


snowfakery/generate_mapping_from_factory.py
```diff
@@ -60,7 +60,7 @@ def generate_record_type_pseudo_tables(summary):
         record_type_pseudo_table.record_type = record_type_name
 
         # copy over the dependencies from the real table
-        for dependency in summary.intertable_dependencies.copy():
+        for dependency in sorted(summary.intertable_dependencies.copy()):
```
davisagli, line 63: There's no reason to use the copy method here, is there?

PR #58, opened 2020-05-26
Date cleanups and features

prescod:
1. Add relativedelta function for date calculations.

2. Simplify rc from parse_date.

3. Allow parsing dates from strings.


prescod:
> Should this be mentioned in the docs?

3602e3c

PR #59, opened 2020-05-26
Improve error reporting

prescod:
More consistent error reporting using a helper.

### DIFF
--- a/snowfakery/data_generator_runtime_dom.py
+++ b/snowfakery/data_generator_runtime_dom.py
@@ -35,7 +35,7 @@ class FieldDefinition(ABC):
 
     @abstractmethod
     def render(self, context: RuntimeContext) -> FieldValue:
-        pass  # noQA
+        pass
 
     @contextmanager
     def exception_handling(self, message: str, *args, **kwargs):
--- a/snowfakery/data_generator_runtime_dom.py
+++ b/snowfakery/data_generator_runtime_dom.py
@@ -287,7 +287,13 @@ class StructuredValue(FieldDefinition):
                     self.filename,
                     self.line_num,
                 )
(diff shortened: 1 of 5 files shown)


snowfakery/data_generator_runtime_dom.py
```diff
@@ -42,6 +42,17 @@ def __init__(self):
...
 
+    @contextmanager
+    def exception_handling(self, message: str, *args, **kwargs):
+        try:
+            yield
+        except DataGenError as e:
+            raise fix_exception(message, self, e)
+        except Exception as e:
+            message = message.format(*args, **kwargs)
+            print(f"{message} : {str(e)}", self.filename, self.line_num)
```
davisagli, line 53: We should use a logger, not print, so that it's possible to configure where the message goes.

prescod: Sorry: It was a debugging print statement that slipped through. Almost want to add a lint rule for those...

snowfakery/data_generator_runtime_dom.py
```diff
@@ -33,14 +33,19 @@ class FieldDefinition(ABC):
...
 
-    def __init__(self):
-        self.definition = None
-        self.filename = None
-        self.line_num = None
-
     @abstractmethod
     def render(self, context: RuntimeContext) -> FieldValue:
-        pass
+        pass  # noQA
```
davisagli, line 38: Were you trying to make coverage.py skip this? Because this isn't the right invocation for that. I'd suggest configuring coverage.py to ignore abstract methods like we have for CumulusCI: https://github.com/SFDO-Tooling/CumulusCI/blob/master/.coveragerc#L8

prescod: Fixed in 9632295 and 85112bf

PR #61, opened 2020-05-28
Syntax overhaul

prescod:
Switch from <<X>> syntax to ${{X}} syntax.

### DIFF
--- a/tests/test_template_funcs.py
+++ b/tests/test_template_funcs.py
@@ -22,8 +22,8 @@ class TestTemplateFuncs(unittest.TestCase):
         - object: Person
           count: 1
           fields:
-            parent: <<reference(daddy).id>>
-            parent2: <<reference(daddy)>>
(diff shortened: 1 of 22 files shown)


tests/test_template_funcs.py
```diff
@@ -155,60 +155,27 @@ def test_conditional_nested(self, write_row):
...
-#             unique_number:
-#                 counter: Person
-#         """
-#         generate(StringIO(yaml), {}, None)
-#         unique_numbers = [
-#             call[1]["unique_number"] for call in write_row_mock.mock_calls
-#         ]
-#         assert unique_numbers == (1, 2, 3, 4, 5)
+    @mock.patch(write_row_path)
+    def test_old_synta(self, write_row):
```
davisagli, line 174:
```suggestion
    def test_old_syntax(self, write_row):
```

PR #62, opened 2020-05-28
Docs in Markdown now.

prescod:


docs/conf.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = "Snowfakery"
+copyright = "2020, Paul Prescod and Salesforce.org Release Engineers"
```
davisagli, line 21: Doesn't Salesforce own the copyright?

docs/index.md
```diff
@@ -0,0 +1,1025 @@
...
+              fields:
+                name: Pets Choice
+                cost: $10
+```
+
+If you’re lost: don't worry! Those are a lot of new concepts at once, and you don't have to understand everything in this example right now. We'll come back to these concepts one by one.
+
+But in case you’re in a hurry: In this case we're creating 3 Person objects. Each one has a name, age, dog and cat. Each dog or cat is an Animal, so we'll get 2 animals per Person or a total of 6. Each animal has a favorite food, so we'll get 6 PetFood objects as well.
+
+<img src='https://quip.com/blob/PJUAAAi0SAB/UReZ0eb4vVO7ebxjxPgLVQ?a=F4goQUMAQOoinOTXGrHGMKYhgKzTZ4BouVaK76LD29Ea' id='PJUACATPCzI' alt='Relationship diagram' width='800' height='190'>Later, we’ll discuss how we could have just 2 PetFood objects which are shared. We’ll also discuss how we could randomly select a Pet species or Food.
```
davisagli, line 143: Are we sure these image links to quip will endure? Should we not include the images in the repository?

docs/conf.py
prescod: Good catch. sphinx-quickstart asks:

> Author name(s): 

Then it outputs a copyright statement!

docs/index.md
prescod: a8f0672

docs/conf.py
prescod: a812325

PR #63, opened 2020-05-30
Great renaming

prescod:
Rename factory->recipe and DOM->Object Model


prescod:
> This renames a module that is imported by CumulusCI. Is the plan to make a new release of Snowfakery, then update CumulusCI to use it?

Yes, that will be the next step.
 
> Are any updates needed to terminology in the docs?

The docs were in another branch when I did the renaming but that's fixed now.

4bda78a

PR #70, opened 2020-06-10
Decode paths to Python strings

prescod:
Occasionally paths are byte-arrays and this decodes them however is "appropriate" for the platform. I ran into this when I was using Snowfakery with FIlesystem2.py


davisagli:
Would a test help ensure that this does in fact do the right thing on different platforms?

PR #97, opened 2020-08-08
Throw plugin exceptions while line# is available

prescod:
Resolve plugins during the parsing process so that the line number and filename are available for error messages.


prescod:
@davisagli : Not an urgent fix but if you have a moment, we should close it out.

setup.py
```diff
@@ -12,9 +11,12 @@
         req for req in dev_requirements_file if not req.startswith("-")
     ]
 
+# get the version into a global variable named "version"
+exec(open("snowfakery/version.py").read())
```
davisagli, line 15: This is not the same approach that we ended up with in https://github.com/SFDO-Tooling/Snowfakery/pull/102. Probably makes sense to merge that one first, then update this branch.

davisagli (changes requested):

davisagli (approved):

PR #102, opened 2020-08-10
Feature/external datasets

prescod:
This is a new feature for merging external data (e.g. a CSV real-address list or a SQL list of classes from a real university) into a Snowfakery process. Docs are here:

https://github.com/SFDO-Tooling/Snowfakery/pull/102/commits/b1c891046df826da9e2e768424c9d75631b8212f


docs/index.md
```diff
@@ -967,6 +967,127 @@ Or:
...
+If you experience performance or memory usage problems with big datasets,
+you might want to shuffle your data into a random order  before using it
+in Snowfakery. Snowfakery's source repository includes a tool called
+`shufflecsv.py` which can do that for CSV files. You feed it a CSV on stdin
+ and it will generate another one on stdout like this:
+
+```bash
+$ python tools/shufflecsv.py < examples/datasets/addresses.csv > examples/datasets/shuffled.csv
+...
+```
```
davisagli, line 1081: Would it make sense to make this a subcommand of `snowfakery` in some fashion? Running it like this requires the user to know where snowfakery's codebase is located. (Not a blocker)

docs/index.md
```diff
@@ -967,6 +967,127 @@ Or:
...
+
+Doing something equivalent for a SQLite or other database
+would involve the clause `order by random()`. This is also
```
davisagli, line 1084: This is pretty handwavy...where would that clause go?

setup.py
```diff
@@ -12,9 +11,12 @@
         req for req in dev_requirements_file if not req.startswith("-")
     ]
 
+# get the version into a global variable named "version"
+exec(open("snowfakery/version.py").read())
```
davisagli, line 15: This doesn't look like it will work on Windows (slash separator, not being careful to close the opened file).

Could we copy CumulusCI's approach and just put the version in a .txt file, partly for consistency and partly to avoid needing to use exec?

snowfakery/data_generator_runtime_object_model.py
```diff
@@ -128,6 +128,10 @@ def name(self) -> str:
             name += " (self.nickname)"
         return name
 
+    @property
+    def id(self):
+        return id(self)
```
davisagli, line 133: Keep in mind that using id() can be problematic if these objects ever need to survive serialization or get swapped out for another object.

snowfakery/plugins.py
```diff
@@ -72,3 +78,22 @@ def lazy(func: Any) -> Callable:
     """A lazy function is one that expects its arguments to be unparsed"""
     func.lazy = True
     return func
+
+
+class PluginResult:
```
davisagli, line 83: docstring (and why is it needed?)

snowfakery/standard_plugins/datasets.py
```diff
@@ -0,0 +1,220 @@
...
+    elif len(tables) > 1:
+        raise Exception(
+            f"Database has multiple tables in it and none was selected: {metadata.tables.keys()}"
+        )
+    if mode == "linear":
+        return SQLDatasetLinearIterator(engine, table)
+    elif mode == "shuffle":
+        return SQLDatasetRandomPermutationIterator(engine, table)
+    else:
+        raise NotImplementedError()
```
davisagli, line 51:
```suggestion
        raise NotImplementedError(f"Unknown mode: {mode}")
```

snowfakery/standard_plugins/datasets.py
```diff
@@ -0,0 +1,220 @@
...
+class SQLDatasetIterator(DatasetIteratorBase):
+    def __init__(self, engine, table):
+        self.connection = engine.connect()
+        self.table = table
+        self.start()
+
+    def start(self):
+        self.results = (
+            PluginResult(dict(row)) for row in self.connection.execute(self.query())
+        )
```
davisagli, line 88: If you anticipate loading large datasets you might want to use https://docs.sqlalchemy.org/en/13/orm/query.html?highlight=yield_per#sqlalchemy.orm.query.Query.yield_per to limit the size of the working set of data fetched from the db

snowfakery/standard_plugins/datasets.py
```diff
@@ -0,0 +1,220 @@
...
+    "Iterator that reads a SQL table in random order"
+
+    def query(self):
+        return select([self.table]).order_by(func.random())
+
+
+class CSVDatasetLinearIterator(DatasetIteratorBase):
+    def __init__(self, datasource: Path):
+        self.datasource = datasource
+        self.file = open(self.datasource)
```
davisagli, line 115: Shouldn't this have an implementation of `close` to close the file?

davisagli: Also, this should specify `newline=''` so that the `csv` module can handle newlines inside of quoted csv fields, and it should specify an encoding if you don't want to get cp1252 on Windows.

snowfakery/standard_plugins/datasets.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    cwd = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(cwd)
+
+
+yaml.SafeDumper.add_representer(quoted_name, Representer.represent_str)
```
davisagli, line 220: This seems like a potentially surprising side effect of importing this plugin. It would be cleaner if snowfakery used its own dumper, perhaps.

tests/conftest.py
```diff
@@ -1,4 +1,24 @@
...
+    def fin():
+        p.stop()
+
+    def row_values(index, value):
+        return mockobj.mock_calls[index][1][1][value]
+
+    request.addfinalizer(fin)
+
+    mockobj.row_values = row_values
+    return mockobj
```
davisagli, line 24: You know you can implement pytest fixtures as a context manager, yield the fixture value, and then do cleanup after the yield statement, right?

tools/shufflecsv.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Shuffle a CSV on stdin to stdout
+#
+# It's much more efficient to shuffle a CSV once instead of every
+# time you run a Snowfakey recipe. So this script does it.
+
+
+def main():
+    firstline = stdin.readline()
+    rows = stdin.readlines()
```
davisagli, line 12: This won't work if there are newlines inside of quoted csv fields.

tests/test_external_datasets.py
```diff
@@ -0,0 +1,241 @@
...
+            nonlocal called
+            called = True
+            return orig_query(self)
+
+        with mock.patch(
+            "snowfakery.standard_plugins.datasets.SQLDatasetRandomPermutationIterator.query",
+            new_query,
+        ):
+            generate(StringIO(yaml), {})
+        assert called
```
davisagli, line 136: I think you could use mock's `wraps` option here: https://stackoverflow.com/a/25608441/212493

davisagli (changes requested):
Sorry for losing track of this one!

docs/index.md
```diff
@@ -967,6 +967,127 @@ Or:
...
+```
+
+`examples/datasets/addresses.csv`
+
+Note that the column numbers come from the first row.
+
+Here is how we can use those rows:
+
+```yaml
+- plugin: snowfakery.standard_plugins.datasets.Dataset
```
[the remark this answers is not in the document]
prescod: It's similar to Python's standard library.

As in Python, it's somewhat arbitrary what's "builtin", versus "standard" versus "3rd party". For example, I use @contextmanager a lot and think it should be builtin. hex() I never use and would prefer it to be in a library.

Maybe the name could be shorter:

- plugin: standard_plugins.datasets.Dataset

Or:

- plugin: snowfakery.datasets.Dataset

Or:

- plugin: datasets.Dataset

docs/index.md
```diff
@@ -967,6 +967,127 @@ Or:
...
+- object: Person
+  count: 10
+  fields:
+    Name:
+      fake: name
+    __address_from_csv:
```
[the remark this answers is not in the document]
prescod: I guess I could rename it to hidden__ or h__? That would be more Salesforce-ish and less Pythonic.

The other special character I can play with is dot.

hidden.address_from_csv

h.address_from_csv

docs/index.md
prescod: That's a good idea. I'll add a ticket to do that.

snowfakery/standard_plugins/datasets.py
prescod: You need that if you're using the ORM but I'm not.

The execute() function returns a ResultProxy:

https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.ResultProxy

Which is "A facade around a DBAPI cursor object."

In general, I suspect we use ORM functions and "the ORM" too much in CumulusCI, by the way. Now that I've worked with SQL Alchemy for 18 months I have a better intuition of when to use the ORM or not. I suspect all of the bulk loading stuff would be simpler if we didn't use any ORM functions but I'm not proposing a big-bang rewrite.

snowfakery/standard_plugins/datasets.py
prescod: Okay I'll add an issue for this. quoted_name objects ARE strings (subclassing) with an extra boolean which seems to me to be relevant only in exceedingly rare situations so I wouldn't call this a blocker. But sure I can fix it up in the next month or so in case Snowfakery is ever embedded in some giant system that round-trips weird tablenames through YAML and has tablenames that differ only in case(?). I agree that it is more polite to try and avoid such a situation even if it seems unlikely.

snowfakery/data_generator_runtime_object_model.py
prescod: I did consider that but the disambiguation is only needed for the lifetime of a recipe execution.

docs/index.md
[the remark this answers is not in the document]
davisagli: I'm not sure it's a significant enough problem to need solving; just reporting how I experienced it.

docs/index.md
```diff
@@ -967,6 +967,127 @@ Or:
...
+```csv
+Number,Street,City,Postcode
+420,Kings Ave,Burnaby,85633
+421,Granville Street,White Rock,85633
+422,Kingsway Road,Richmond,85633
+```
+
+`examples/datasets/addresses.csv`
+
+Note that the column numbers come from the first row.
```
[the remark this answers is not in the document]
prescod: 744f512a5de

docs/index.md
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/102/commits/744f512a5de7cf306e7d5e94815459fd8af82263#diff-1a523bd9fa0dbf998008b37579210e12R1089

snowfakery/__init__.py
```diff
@@ -1 +1,12 @@
...
 from .plugins import SnowfakeryPlugin, lazy  # noqa
+
+
+# TODO: when Python 3.6 is irrelevant, make this lazy:
+
+# https://www.python.org/dev/peps/pep-0562/
+
+version_file = Path(__file__).parent / "version.txt"
+with version_file.open() as f:
+    version = f.read().strip()
```
davisagli, line 12: This is fine, but you could have done `version = version_file.read_text().strip()`

setup.py
prescod: I did change it but just for interest's sake:

Slashes do work on Windows. And Windows will also close files automatically. The issue is actually that Windows hates to have an open file that is not in a directory whereas Unix doesn't care. So in Windows you must be careful to close files before deleting directories and therefore manual closing encourages you to think through the order of resource cleanup to make sure that the file handle goes away before the directory.

tests/conftest.py
```diff
@@ -1,4 +1,23 @@
...
+
+    def row_values(index, value):
+        return mockobj.mock_calls[index][1][1][value]
+
+    mockobj.row_values = row_values
+    yield mockobj
+
+    p.stop()
+
+    return
```
davisagli, line 23: The return is a no-op, isn't it?

You also could use patch as a context manager and avoid the explicit calls to start/stop. That would help avoid the bug here where stop doesn't get called if the test raises an exception

tests/test_external_datasets.py
prescod: The SO recipe requires me to create a SQLDatasetRandomPermutationIterator which takes two arguments which I'd need to create or mock.

And the recipe has an issue described in the comments which is that the method which is called is bound to a different to the dummy Test object rather than the one created in the code. So i guess it would be querying the wrong database?

And a naive wraps() is not sufficient because Mock objects can't be used as unbound methods. Python won't bind them because it only binds "real" functions. Other callable objects don't get bound.

snowfakery/standard_plugins/datasets.py
prescod: https://github.com/SFDO-Tooling/Snowfakery/issues/132

tests/conftest.py
prescod: 744b9bb

davisagli (approved):

PR #106, opened 2020-08-14
Salesforce Example Files

prescod:
Some example files for Salesforce users including flows and scripts showing how to use Snowfakery from CCI.


examples/salesforce/scripts/test_opportunities_npsp.sh
```diff
@@ -0,0 +1 @@
+cci task run generate_and_load_from_yaml -o generator_yaml examples/npsp/Opportunity_npsp.recipe.yml -o num_records 300 -o num_records_tablename Opportunity --org qa
```
davisagli, line 1: What's the reason for the recipe files mentioning these scripts which just run a one-liner, instead of just including the one-liner in the comment in the recipe? I think the latter would be preferable in order to avoid the use of shell scripts, which don't port to Windows as easily.

prescod: Fixed:

https://github.com/SFDO-Tooling/Snowfakery/pull/106/commits/b8eb9e33d317e732af485a6825b1f74f9fd6793a

davisagli: @prescod are these shell scripts still used?

davisagli (approved):
@prescod I don't have much basis for reviewing the actual example data structure here, so mine was a pretty cursory review.

PR #113, opened 2020-08-20
Allow forward references to nicknames.

prescod:
Support Snowfakery files like this:

```
- object: A
  fields:
    B:
      reference: Bob. # referenced before it is defined.
- object: B
  nickname: Bob
  fields:
    A:
      reference: A
```

A new object is created called EmptySlot which is allocated each time the recipe is instantiated, analogous to local variable slots in Python, or NamedTuple slots.

When a slot is allocated, it grabs a database ID just as an ObjectRow would. Then, when the object is created, it uses that ID.

If the ID is never used by any ObjectRow, that will cause a reportable error.

### DIFF
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -372,7 +372,7 @@ class RuntimeContext:
 
     def check_if_finished(self):
         "Have we iterated over the script enough times?"
-        # TODO: check that every forward reference was resolved
+        # check that every forward reference was resolved
         self.interpreter.globals.check_slots_filled()
         return self.interpreter.finished_checker.check_if_finished(
             self.interpreter.globals.id_manager
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -179,7 +179,7 @@ class Globals(yaml.YAMLObject):
         name_slots: Mapping[str, str] = (),
     ):
         # list starts empty and is filled. Survives continuations.
-        self.named_objects = {}
+        self.nicknamed_objects = {}
 
         self.id_manager = IdManager(start_ids)
         self.last_seen_obj_of_type = {}
@@ -191,17 +191,22 @@ class Globals(yaml.YAMLObject):
     def register_object(self, obj, nickname: str = None):
         """Register an object for lookup by object type and (optionally) Nickname"""
         if nickname:
-            self.named_objects[nickname] = obj
+            self.nicknamed_objects[nickname] = obj
         self.last_seen_obj_of_type[obj._tablename] = obj
 
     @property
     def object_names(self):
         """The globally named objects"""
-        return {**self.named_slots, **self.named_objects, **self.last_seen_obj_of_type}
+        # the order is important: later overrides earlie
+        return {
+            **self.named_slots,  # potential forward or backwards references
+            **self.nicknamed_objects,  # nicknames that have been fulfilled
+            **self.last_seen_obj_of_type,  # tablenames that have been fulfilled
+        }
 
     def generate_id_for_nickname(self, nickname: str):
-        slot = self.named_slots[nickname]
-        if slot.status == SlotState.ALLOCATED:
+        slot = self.named_slots.get(nickname)
+        if slot and slot.status == SlotState.ALLOCATED:
             return slot.consume_slot()
 
     def register_intertable_reference(
@@ -310,11 +315,20 @@ class Interpreter:
         }
         self.finished_checker = FinishedChecker(start_ids, stopping_criteria)
         self.faker_template_library = FakerTemplateLibrary(faker_providers)
+
+        # TODO: clean this duplicated code up
         name_slots = {
             template.nickname: template.tablename
             for template in templates
             if template.nickname
         }
+
+        # table names are sort of nicknames for themselves too, because
+        # you can refer to them.
+        name_slots.update(
+            {template.tablename: template.tablename for template in templates}
+        )
+        # print("XXXX", name_slots, globals)
 
         self.globals = globals or Globals(name_slots=name_slots)
 
@@ -381,8 +395,16 @@ class RuntimeContext:
     def generate_id(self, nickname: Optional[str]):
         "Generate a unique ID for this object"
         rc = None
+        # check if an ID has already been assigned based on the nickname
+        # (due to a forward reference)
         if nickname:
             rc = self.interpreter.globals.generate_id_for_nickname(nickname)
+        # check if an ID has already been assigned based on the tablenmae.
+        # (due to a forward reference)
+        rc = rc or self.interpreter.globals.generate_id_for_nickname(
+            self.current_table_name
+        )
+        # othewise just create a new one
         rc = rc or self.interpreter.globals.id_manager.generate_id(
             self.current_table_name
         )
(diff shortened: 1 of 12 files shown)


snowfakery/output_streams.py
```diff
@@ -68,7 +68,7 @@ def flatten(
         sourcetable: str,
         fieldname: str,
         source_row_dict: Dict,
-        target_object_row: ObjectRow,
+        target_object_row: [ObjectRow, NicknameSlot],
```
davisagli, line 71: I'm not familiar with this syntax in type annotations. Does it actually work, or is it supposed to be `Union[ObjectRow, NicknameSlot]`? There are a few repetitions below too.

snowfakery/data_generator_runtime.py
```diff
@@ -119,6 +121,45 @@ class Dependency(NamedTuple):
 )
 
 
+class SlotState(Enum):
```
davisagli, line 124: docstring

snowfakery/data_generator_runtime.py
```diff
@@ -119,6 +121,45 @@ class Dependency(NamedTuple):
...
+    UNUSED = auto()
+    ALLOCATED = auto()
+    CONSUMED = auto()
+
+
+class NicknameSlot:
```
davisagli, line 130: docstring

snowfakery/data_generator_runtime.py
```diff
@@ -119,6 +121,45 @@ class Dependency(NamedTuple):
...
+
+    def __init__(self, tablename, id_manager):
+        self._tablename = tablename
+        self.id_manager = id_manager
+
+    @property
+    def id(self):
+        if self.allocated_id is None:
+            self.allocated_id = self.id_manager.generate_id(self._tablename)
+        return self.allocated_id
```
davisagli, line 143: I'm looking forward to when we can use https://docs.python.org/3/library/functools.html#functools.cached_property for this sort of memoization (it was added in Python 3.8)

snowfakery/data_generator_runtime.py
```diff
@@ -119,6 +121,45 @@ class Dependency(NamedTuple):
...
+        return rc
+
+    @property
+    def status(self):
+        if self.allocated_id is None:
+            return SlotState.UNUSED
+        elif isinstance(self.allocated_id, int):
+            return SlotState.ALLOCATED
+        elif self.allocated_id == SlotState.CONSUMED:
+            return SlotState.CONSUMED
```
davisagli, line 157: Isn't there a risk of collision between the SlotState enum integers and actual object ids?

snowfakery/data_generator_runtime.py
```diff
@@ -131,12 +172,21 @@ class Globals(yaml.YAMLObject):
     yaml_dumper = yaml.SafeDumper
     yaml_tag = "!snowfakery_globals"
 
-    def __init__(self, start_ids: Optional[Dict[str, id]] = None, today: date = None):
+    def __init__(
+        self,
+        start_ids: Optional[Dict[str, id]] = None,
+        today: date = None,
+        name_slots: Mapping[str, str] = (),
```
davisagli, line 179: A mapping type that defaults to a tuple?

snowfakery/data_generator_runtime.py
```diff
@@ -156,16 +211,40 @@ def register_intertable_reference(
...
+            if slot.status == SlotState.ALLOCATED
+        ]
+        if len(not_filled) > 1:
+            raise DataGenNameError(
+                f"References not fulfilled: {','.join(not_filled)}", None, None
+            )
+        elif not_filled:
+            raise DataGenNameError(
+                f"Reference not fulfilled: {not_filled[0]}", None, None
+            )
```
davisagli, line 234: Why do you need a special case for length 1?

snowfakery/data_generator_runtime.py
```diff
@@ -241,6 +326,21 @@ def __init__(
...
+    def loop_over_templates(self, runtimecontext, continuing):
+        finished = False
+        while not finished:
+            self.loop_over_templates_once(runtimecontext, continuing)
+            finished = runtimecontext.check_if_finished()
+            continuing = True
+
+    def loop_over_templates_once(self, runtimecontext, continuing):
+        for template in self.templates:
+            should_skip = template.just_once and continuing is True
```
davisagli, line 340:
```suggestion
            should_skip = template.just_once and continuing
```

snowfakery/data_generator_runtime.py
```diff
@@ -272,13 +372,21 @@ def __init__(
 
     def check_if_finished(self):
         "Have we iterated over the script enough times?"
+        # TODO: check that every forward reference was resolved
```
davisagli, line 375:
```suggestion
        # check that every forward reference was resolved
```

snowfakery/output_streams.py
prescod: That was a typo and for some reason TypeGuard didn't catch it. Fixed now.

snowfakery/data_generator_runtime.py
prescod: No, the enums are pretty robust from a type perspective.

```
>>> SlotState.UNUSED
<SlotState.UNUSED: 1>
>>> SlotState.UNUSED==1
False
>>> int(SlotState.UNUSED)==1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a number, not 'SlotState'
>>> SlotState.UNUSED.value == 1
True
```

snowfakery/data_generator_runtime.py
prescod: It's immutable and it responds to iter, __contains__. There is no "empty immutable mapping type." It actually works nicely although its true that mypy doesn't like it. Slightly frustrating because it is more mapping-like than None, which mypy does allow. (of course None has all sorts of special processing around it)

I guess I'll avoid that convention in the future to minimize future rework when we use MyPy.

snowfakery/data_generator_runtime.py
prescod: 7b57423

snowfakery/data_generator_runtime.py
prescod: 7b57423

snowfakery/data_generator_runtime.py
prescod: 7b57423

snowfakery/data_generator_runtime.py
prescod: https://github.com/SFDO-Tooling/Snowfakery/commit/7b57423dafe18b0e5d76f02ec3d3eb453f4a4ed2#diff-643cb5ba59a4a93cb618c80e5ce3fc18R188

snowfakery/data_generator_runtime.py
prescod: https://github.com/SFDO-Tooling/Snowfakery/commit/7b57423dafe18b0e5d76f02ec3d3eb453f4a4ed2#diff-643cb5ba59a4a93cb618c80e5ce3fc18R135

snowfakery/data_generator_runtime.py
prescod: https://github.com/SFDO-Tooling/Snowfakery/commit/7b57423dafe18b0e5d76f02ec3d3eb453f4a4ed2#diff-643cb5ba59a4a93cb618c80e5ce3fc18R125

snowfakery/output_streams.py
prescod: https://github.com/SFDO-Tooling/Snowfakery/commit/29e017095a45b903743fab85bf2ea53b174c8040#diff-bd3afc575a9fee6d617119045da1b50eR71

prescod:
> This looks like a handy improvement. Does it warrant any updates to the docs?

https://github.com/SFDO-Tooling/Snowfakery/commit/29e017095a45b903743fab85bf2ea53b174c8040#diff-1a523bd9fa0dbf998008b37579210e12R415

PR #114, opened 2020-08-20
Check with Faker before dateutil.parser.parse

### LINKED ISSUE #42: Using dateutil.parser prevents some helpful date offsets
While generating a date field with the `date_between` function block, I can use a min/max offset value of days, like the documentation says:

```
SomeFieldName:
  date_between:
    min: -120d
    max: -90d
```

but I can't use *month* as the unit for this offset, even though [faker's date_time provider](https://faker.readthedocs.io/en/stable/providers/faker.providers.date_time.html#faker.providers.date_time.Provider.date_between) [supports it](https://github.com/joke2k/faker/blob/f874509ad8eb551ca70c5a69d9b41aa356d8d1bc/faker/providers/date_time/__init__.py#L60).  If I try:

```
SomeFieldName:
  date_between:
    min: -4M
    max: -3M
```

Then I silently get the wrong result: for min, a datetime with today's date, and 00:04:00 as the time; for max,  a datetime with today's date, and 00:03:00 as the time.

It looks like this is because snowfakery sends the value through `dateutil.parser.parse()` first (in template_funcs:parse_date()).  For a days-back value like `-120d`, `dateutil.parser.ParserError` is raised so `-120d` gets passed to faker and everything works.  For a months-back value like `-3M`, dateutil.parser doesn't raise ParserError and returns an unexpected datetime.

Expected behavior: snowfakery's date values support all of faker's formats.  Note that snowfakery's documentation says values like `-4m` are supported for months, but this has the same output as described for `-4M` above, and if faker's formats were followed, `-4m` would be *four minutes ago*.

prescod:
Fix #42

### DIFF
--- a/tests/test_faker.py
+++ b/tests/test_faker.py
@@ -1,6 +1,7 @@
 from io import StringIO
 import unittest
 from unittest import mock
+from datetime import date
 
 from snowfakery.data_generator import generate
 
@@ -94,3 +95,18 @@ class TestFaker(unittest.TestCase):
         """
         generate(StringIO(yaml), {}, None)
         assert row_values(write_row_mock, 0, "date") is None
+
+    @mock.patch(write_row_path)
+    def test_months_past(self, write_row_mock):
+        yaml = """
+        - object: A
+          fields:
+            date:
+              date_between:
+                start_date: -4M
+                end_date: -3M
+        """
+        generate(StringIO(yaml), {}, None)
+        the_date = row_values(write_row_mock, 0, "date")
+        assert (date.today() - the_date).days > 80
+        assert (date.today() - the_date).days < 130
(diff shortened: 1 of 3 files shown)


prescod:
@adamlincoln

tests/test_faker.py
```diff
@@ -94,3 +95,18 @@ def test_non_overlapping_dates(self, write_row_mock):
...
+          fields:
+            date:
+              date_between:
+                start_date: -4M
+                end_date: -3M
+        """
+        generate(StringIO(yaml), {}, None)
+        the_date = row_values(write_row_mock, 0, "date")
+        assert (date.today() - the_date).days > 80
+        assert (date.today() - the_date).days < 130
```
davisagli, line 112: I like the ~10 day "I don't wanna think about variation in month lengths" fudge factor. :)

prescod:
Doc update: https://github.com/SFDO-Tooling/Snowfakery/pull/114/files?file-filters%5B%5D=.py#diff-1a523bd9fa0dbf998008b37579210e12

PR #119, opened 2020-08-26
Advance dependencies per dependabot

prescod:


davisagli:
@prescod there's a conflict...

PR #120, opened 2020-08-27
Macros can include other macros

prescod:
For example:

```
- macro: foo
  fields:
    foobar: FOOBAR

- macro: bar
  include: foo
  fields:
    barbar: BARBAR

- object: Bar
  include: bar
```

Generates:

```
Bar(id=1, barbar=BARBAR, foobar=FOOBAR)
```


davisagli (changes requested):
Docs?

prescod:
This PR changes the implementation to match the docs.

https://snowfakery.readthedocs.io/en/docs/#macros

"Macros can themselves include other macros."

davisagli (approved):

PR #129, opened 2020-09-09
Week 37 dependency updates

Br4nd0R:


requirements_dev.txt
```diff
@@ -1,9 +1,9 @@
 -r requirements.txt
-black==19.10b0
+black==20.8b1
```
davisagli, line 2: Make sure you update the `rev` for the black pre-commit hook to match, in .pre-commit-config.yaml. And then you'll need to run black on everything to pass the pre-commit hook.

davisagli (approved):

PR #133, opened 2020-09-12
Parse errors should not cause JSON with close-bracket

### LINKED ISSUE #131: Output file written when recipe fails to parse
I noticed that when Snowfakery is given an invalid recipe, and at least when it is set to create JSON output it generates a file consisting only of a single `]`.

Reproduction steps:
1. Take a valid file, and introduce a simple error that will stop the parser (it looks like any error will do).
2. Run snowfakery directly: `snowfakery.exe .\datasets\gen_objects.yml --output-format json -o .\datasets\objects.json`
3. Get expected error that the input was invalid (in my case a field name out of place):
```
An error occurred. If you would like to see a Python traceback, use the --debug-internals option.
Error: Unexpected key: name
 near \blah\blah\datasets\gen_objects.yml:56
```
4. Check the new object and it contains the closing bracket but nothing else.  I expected that the output file would be untouched (either still missing or still the previous version), and I would understand blank as a valid result.

OS: Windows 10
snowfakery, version 1.0.1 installed via pipx
OS Python 3.8.5

prescod:
Fixes #131

### DIFF
--- a/snowfakery/output_streams.py
+++ b/snowfakery/output_streams.py
@@ -229,7 +229,8 @@ class JSONOutputStream(FileOutputStream):
         self.stream.write(json.dumps(values))
 
     def close(self) -> None:
-        self.stream.write("]\n")
+        if not self.first_row:
+            self.stream.write("]\n")
         super().close()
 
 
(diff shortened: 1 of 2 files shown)


davisagli (changes requested):
Could you add a test for the scenario this fixes?

davisagli (approved):

PR #141, opened 2020-09-21
Continuation file changes

prescod:
1. Reduce the dependency on custom tags so that renaming classes won't cause problems for very long-running jobs. (although other format changes could of course do so)

2. Use a custom dumper so my customizations won't leak into CCI from Snowfakery

ObjectRows and everything they can contain are still managed with tags and custom representers because YAML does a good job of managing the inter-object references, heterogenous type conversions and recursion and I didn't see much benefit in re-inventing all of those features.


davisagli:
@prescod can you resolve the conflicts?

snowfakery/data_generator.py
```diff
@@ -73,13 +77,17 @@ def merge_options(option_definitions: List, user_options: Mapping) -> Tuple[Dict
...
-def load_continuation_yaml(continuation_file: TextIO):
+def load_continuation_yaml(continuation_file: FileLike):
     """Load a continuation file from YAML."""
-    return safe_load(continuation_file)
+    return hydrate(Globals, yaml.safe_load(continuation_file))
 
 
-def save_continuation_yaml(continuation_data: Globals, continuation_file: TextIO):
-    safe_dump(continuation_data, continuation_file)
+def save_continuation_yaml(continuation_data: Globals, continuation_file: FileLike):
```
davisagli, line 85: docstring

snowfakery/data_generator_runtime.py
```diff
@@ -93,18 +90,17 @@ def generate_id(self, table_name: str) -> int:
...
 
+    def __getstate__(self):
+        return {"last_used_ids": dict(self.last_used_ids)}
+
     def __setstate__(self, state):
         self.last_used_ids = defaultdict(lambda: 0, state["last_used_ids"])
 
 
-yaml.SafeDumper.add_representer(defaultdict, yaml.SafeDumper.represent_dict)
+SnowfakeryDumper.add_representer(defaultdict, SnowfakeryDumper.represent_dict)
```
davisagli, line 100: Would it make sense to centralize these where SnowfakeryDumper is defined instead of scattered throughout the code?

snowfakery/data_generator_runtime.py
```diff
@@ -235,22 +227,44 @@ def check_slots_filled(self):
...
+            "today": today,
+            "intertable_dependencies": intertable_dependencies,
+        }
 
-    def __setstate__(self, state):
-        self.last_seen_obj_of_type = {}
-        for slot, value in state.items():
-            setattr(self, slot, value)
-        self.intertable_dependencies = set()
+    def __setstate__(self, dict):
```
davisagli, line 255: Call it d instead of dict to avoid shadowing a builtin

davisagli (changes requested):
This is not backwards-compatible with existing continuation files, right? We should be sure to call that out in the release notes.

snowfakery/data_generator_runtime.py
prescod: I thought about it, but felt like it would end up being a sort of hub-and-spoke which might cause circular dependencies and otherwise-unrelated imports centralized in one file. defaultdict is a bit of a special case because its a standard function but for example PluginResult is a "concept" that makes the most sense in the context of the plugins module. There are only 4 of them though, so I don't feel strongly one way or the other.

snowfakery/data_generator_runtime.py
prescod: 49fe14d

snowfakery/data_generator.py
prescod: 49fe14d

davisagli (approved):

davisagli (dismissed):

prescod:
Sorry: every time I merge master, pycoverage gets confused and claims I'm reducing coverage.

davisagli (approved):

PR #149, opened 2020-10-02
Fixes to Random Choice Docs, and trim trailing white space

acrosman:
The docs contained a couple of references to `Random_Choice` syntax that was either wrong or out of date.  This commit makes them all consistent.  It also trims off trailing white space on the end of a number of lines.


prescod:
So I finally got around to investigation of this. I wanted to know what effort it would take to also support the short-form. It seems not too bad based on my experiments. What do you think of doing that instead @acrosman ?

The most common version of it would look like this:

```
        - object : A
          fields:
            b:
                random_choice:
                    A: 80%
                    B: 10%
                    C: 10%
```

And a more complex version would look like this:

```
        - object : A
          fields:
            __a_percent:
                random_number:
                    min: 40
                    max: 60
            b:
                random_choice:
                    A: ${{__a_percent}}%
                    B: ${{__a_percent / 2}}%
                    C: ${{100 - (__a_percent * 1.5)}}%
```

The only case where the very verbose syntax MIGHT still be valuable is if the enum values are the computed value rather than the keys, but I don't have a use-case in mind where that would be helpful. Either way, the "current" syntax would be supported for backwards compatibility for quite a while.

acrosman:
I had a personal preference for the short form of the syntax -- that's how I ran into it not working. So I'm very strongly in support of pushing in the direction.  Other than ease of parsing (and now backward compatibility) I'm not sure what upside there would be to the this structure:
```      
    random_choice:
        - choice:
            probability: 90%
            pick: Closed Won
```

The syntax with the __ prefixed field name is interesting.  Looks like you are proposing (using? I haven't noticed this in the docs before but that doesn't mean this isn't there) a private field that's creating a value that can be used as a variable but doesn't end up in the generated data file.  Makes me wonder if a parallel construct for a private object would make sense for holding variables generated at start up would be useful (although it may not have any advantage over the existing variables in the recipes).

prescod:
@acrosman With respect to the question of why you might want the longer form of random_choice, I've now documented that here:

https://github.com/SFDO-Tooling/Snowfakery/pull/172/files#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6R452

With respect to hidden fields, those are documented here:

https://snowfakery.readthedocs.io/en/docs/#advanced-features

acrosman:
@prescod that makes a lot of sense. 

With both styles staying it might be useful to add more text to the random_choice docs section to fully describe what each style can do. Then leave the bulk of the examples in the doc in the short form unless the longer form features are needed (making this PR largely incorrect).

davisagli:
Closing this since @prescod added support for the short syntax.

PR #156, opened 2020-10-14
Check file extension on CLI

prescod:
Give a proper error message if user tries to get Snowfakery to infer the filetype of from file extension it doesn't know about.


snowfakery/cli.py
```diff
@@ -212,8 +212,12 @@ def configure_output_stream(
                 output_streams.append(JSONOutputStream(path))
             elif format == "txt":
                 output_streams.append(DebugOutputStream(path))
-            else:
+            elif format in file_extensions:
                 output_streams.append(ImageOutputStream(path, format))
+            else:
+                raise click.ClickException(
+                    f"Unknown format or file extension: {format}"
+                )
```
davisagli, line 220: missing test coverage for both these blocks

davisagli (approved):

davisagli:
@prescod I removed the coveralls commit status from the list of commit statuses that are required to merge. I've never quite understood what it compares coverage to and it seems to draw the wrong conclusion sometimes.

It's up to us to click through the coveralls report and check if the modified code is covered. It'll get simpler if we can achieve 100% coverage and just make the build fail if any coverage is missing.

PR #157, opened 2020-10-14
PluginsResults with complex state state can implement "simplify"

prescod:
Outputstreams will call "simplify" for PluginResults.

Example is in test_custom_plugins_and_providers

Lazy plugins can either ask for their parameters to be simplified
by calling context.evaluate() or they can use
context.evaluate_raw() if they don't care at all about the data
of their parameters (higher order function-type stuff).

A few type definitions moved around to permit code reuse without
circular references.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1143,6 +1143,10 @@ expose a namespace that other code can access through dot-notation. PluginResult
 initialized with either a dict or an object that exposes the namespace through Python 
 getattr().
 
+If your plugin generates some special kind of data value which should be serializable
+as a primitive type (usually a string), subclass PluginResult and add a `simplify`
+method to your PluginResult. That method should return a Python primitive value.
+
 In the rare event that a plugin has a function which need its arguments to be passed to it unevaluated, for later (perhaps conditional) evaluation, you can use the `@snowfakery.lazy decorator`. Then you can evaluate the arguments with `self.context.evaluate()`. 
 
 For example:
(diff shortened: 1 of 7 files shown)


docs/index.md
```diff
@@ -1179,6 +1183,11 @@ This would output an `OBJ` row with values:
   {'id': 1, 'some_value': 'abc : abc', 'some_value_2': '1 : 2'})
 ```
 
+Occasionally you might write a plugin which needs to evaluate its
+parameters lazily but doesn't care about the internals of the values
+because it just returns it to some parent context. In that case,
+use `context.evaluate_raw` instead of `context.evaluate`.
+
 ## Using Snowfakery within CumulusC
```
davisagli, line 1191:
```suggestion
## Using Snowfakery within CumulusCI
```

snowfakery/plugins.py
```diff
@@ -74,9 +80,20 @@ def context_vars(self):
...
 
-    def evaluate(self, field_definition):
+    def evaluate_raw(self, field_definition):
+        """Evaluate the contents of a field definition"""
         return field_definition.render(self.interpreter.current_context)
 
+    def evaluate(self, field_definition):
+        """Evaluate the contents of a field definition and simplify to a primitive value."""
+        rc = self.evaluate_raw(field_definition)
+        if isinstance(rc, Scalar.__args__):
```
davisagli, line 90: Will this be a problem if rc is None, or is that not possible?

davisagli (commented):
Not sure I entirely understand what's motivating this, but I assume you know what you're doing.

snowfakery/plugins.py
prescod: The Scalar datatype does include None and None is basically a valid value in most contexts of Snowfakery. Outputstreams can output "null" or ,, or whatever is appropriate.

davisagli (approved):

PR #161, opened 2020-10-20
Fix exception raise in the case of a bad plugin.

prescod:
Improve incorrect error handling.


snowfakery/plugins.py
```diff
@@ -92,7 +92,9 @@ def evaluate(self, field_definition):
         elif hasattr(rc, "simplify"):
             return rc.simplify()
         else:
-            raise f"Cannot simplify {field_definition}. Perhaps should have used evaluate_raw?"
+            raise exc.DataGenError(
+                f"Cannot simplify {field_definition}. Perhaps should have used evaluate_raw?"
+            )
```
davisagli, line 97: This has a bug which you will discover when you add test coverage.

davisagli (changes requested):

PR #162, opened 2020-10-23
SF can now generate SQL text as an output format compatible with CCI's data loader

prescod:
CCI now generates SQL text as an output-format.


docs/index.md
```diff
@@ -154,10 +154,19 @@ Snowfakery builds on a tool called SQLAlchemy, so it gets a variety of database
...
 
-Snowfakery can also output JSON, directories of CSV and object diagrams.
+Snowfakery can also output JSON, SQL, directories of CSV and object diagrams.
 
 CSV output goes to a directory with one CSV file per table and a JSON manifest file in the [csvw](https://www.w3.org/TR/tabular-data-primer/) format.
 
+The complete list of file-based (as opposed to database-based) formats are:
+
+- JSON - a custom JSON dialect
+- TXT - debugging-style output)
```
davisagli, line 164:
```suggestion
- TXT - debugging-style output
```

docs/index.md
```diff
@@ -154,10 +154,19 @@ Snowfakery builds on a tool called SQLAlchemy, so it gets a variety of database
...
+- CSV - a directory of CSV files plus a csvw file
+- SQL - a bulk import SQL file
```
davisagli, line 166: "bulk import"?

snowfakery/output_streams.py
```diff
@@ -117,10 +118,10 @@ def close(self) -> Optional[Sequence[str]]:
 
         Return a list of messages to print out.
         """
-        pass
+        super().close()
 
 
-class FileOutputStream(OutputStream):
+class SmartStream:
```
davisagli, line 124: docstring? what's so smart about it?

snowfakery/output_streams.py
```diff
@@ -132,20 +133,27 @@ def __init__(self, stream_or_path=None):
...
 
+class FileOutputStream(OutputStream, SmartStream):
+    pass
+
+
 class DebugOutputStream(FileOutputStream):
     def write_single_row(self, tablename: str, row: Dict) -> None:
         values = ", ".join([f"{key}={value}" for key, value in row.items()])
-        print(f"{tablename}({values})", file=self.stream)
+        self.write(f"{tablename}({values})\n")
```
davisagli, line 156: Should we use \r\n on Windows?

snowfakery/output_streams.py
```diff
@@ -216,20 +224,20 @@ def __init__(self, file):
...
         values = {"_table": tablename, **row}
-        self.stream.write(json.dumps(values))
+        self.write(json.dumps(values))
 
     def close(self) -> Optional[Sequence[str]]:
         if not self.first_row:
-            self.stream.write("]\n")
-        return super().close()
+            self.write("]\n")
+        super().close()
```
davisagli, line 237: Shouldn't this return the return value from the super call?

docs/index.md
prescod: Would "bulk insert" work for you? During the editing process for my blog post the writer pointed out that many people think of SQL as a "query language" and therefore it helps to clarify to be more specific.

snowfakery/output_streams.py
prescod: I think Windows tools support \n these days and I'd rather not pollute the world with more bloated files. By coincidence, I posted this to Facebook this weekend:

"I was debugging something recently and I noticed something horrifying which had never risen to my conscious observation before.

Every time we load a web page or send almost any message across the Internet, the message is broken into many lines. Each line ends with two characters: carriage return, line feed. Why two characters instead of one?

Those two characters are used for compatibility with older protocols from the early Internet which used the two characters. These protocols were compatible with older operating systems like DOS, CP/M, VMS.
These operating systems were trying to be compatible with teletype machines which are more or less electrical long-distance typewriters.

It is only a slight exaggeration to say that many billions (probably trillions) of extra, completely useless, characters are sent over the Internet every day to be compatible with glorified typewriters which were standardized in the 1960s."

(But QUIC will fix it eventually.)

docs/index.md
prescod: fc94b40

snowfakery/output_streams.py
prescod: 38c9eacf

snowfakery/output_streams.py
prescod: 38c9eacf

docs/index.md
davisagli: Yes using "insert" is clearer, thanks

snowfakery/output_streams.py
```diff
@@ -117,10 +120,12 @@ def close(self) -> Optional[Sequence[str]]:
 
         Return a list of messages to print out.
         """
-        pass
+        return super().close()
 
 
-class FileOutputStream(OutputStream):
+class SmartStream:
+    """Common code for managing stream/file opening/closing"""
```
davisagli, line 127:
```suggestion
class SmartStream:
    """Common code for managing stream/file opening/closing
    
    Expects to be initialized with either a file-like object with a `write` method,
    or a path (str or pathlib.Path) that can be opened using `open()`
    """
```

snowfakery/output_streams.py
```diff
@@ -132,20 +137,31 @@ def __init__(self, stream_or_path=None):
...
+    def close(self):
         if self.owns_stream:
             self.stream.close()
             return [f"Generated {self.stream.name}"]
 
 
+class FileOutputStream(OutputStream, SmartStream):
+    """Base class for all file/stream-based OutputStreams"""
+
+    pass
```
davisagli, line 156:
```suggestion
```

(This one's a matter of taste; feel free to disregard)

snowfakery/output_streams.py
davisagli: I'm fine with just using `\n` until we have a concrete reason to do something more complicated.

snowfakery/output_streams.py
```diff
@@ -132,20 +137,31 @@ def __init__(self, stream_or_path=None):
...
+
+
 class DebugOutputStream(FileOutputStream):
+    """Simplied output for debugging Snowfakery files."""
```
davisagli, line 160:
```suggestion
    """Simplified output for debugging Snowfakery files."""
```

davisagli (approved):

PR #164, opened 2020-10-30
Document and improve usability of target-number feature

prescod:
The target-number command line was not documented. Now it is.

While documenting it, I thought it would be better if it read in the same order as English: "50 Account"(s). Both orders are supported in the code, however.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -886,9 +886,7 @@ Options:
 
 ### Scaling up recipe execution
 
-From the command line you can control how many rows a recipe generates by telling
-Snowfakery how many times to execute it. You do this by specifying a "target count"
-and a "target tablename", like this:
+From the command line you can control how many rows a recipe generates. You do this by specifying a "target count" and a "target tablename", like this:
 
 ```bash
 snowfakery accounts.yml --target-number 1000 Account
--- a/docs/index.md
+++ b/docs/index.md
@@ -884,7 +884,31 @@ Options:
   
 ```
 
-## CSV Output
+### Scaling up recipe execution
+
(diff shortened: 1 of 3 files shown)


Br4nd0R (dismissed):
Looks good!

docs/index.md
```diff
@@ -884,7 +884,31 @@ Options:
   
 ```
 
-## CSV Output
+### Scaling up recipe execution
+
+From the command line you can control how many rows a recipe generates by telling
+Snowfakery how many times to execute it. You do this by specifying a "target count"
```
davisagli, line 890: Based on the explanation below, "how many times to execute it" is a bit misleading.

snowfakery/cli.py
```diff
@@ -267,6 +283,18 @@ def validate_options(
         )
 
 
+def stopping_criteria_from_target_number(target_number):
+    "Deconstruct a tuple of 'str number' or 'number str' and make a StoppingCriteria"
+
+    # 'number str' is the official format so the other one can be deprecated one day.
```
davisagli, line 289: Isn't putting them in a canonical order already accomplished by int_string_tuple?

docs/index.md
prescod: 99d3fc3

snowfakery/cli.py
prescod: I'll merge the embedding API next.

https://github.com/SFDO-Tooling/Snowfakery/compare/feature/embedding-api?expand=1

davisagli (approved):

PR #165, opened 2020-10-30
Do not serialize deep object references to continuation file

prescod:
Do not serialize deep object references to continuation file. They sometimes cause reference loops which YAML cannot (in this case) load.

### DIFF
--- a/tests/test_restartability.py
+++ b/tests/test_restartability.py
@@ -44,6 +44,8 @@ class TestRestart(unittest.TestCase):
     def test_faker_dates_work(self, write_row):
         yaml = """
             - object: foo
+              nickname: Blah
+              just_once: True
               count: 50
               fields:
                 a_date:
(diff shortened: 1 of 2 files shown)


tests/test_restartability.py
```diff
@@ -54,3 +56,26 @@ def test_faker_dates_work(self, write_row):
...
+                                TheParent
+            """
+        continuation_file = StringIO()
+        generate(StringIO(yaml_data), generate_continuation_file=continuation_file)
+        continuation_yaml = continuation_file.getvalue()
+        print(continuation_yaml)
+        generate(
+            StringIO(yaml_data),
+            continuation_file=StringIO(continuation_yaml),
+        )
```
davisagli, line 81: @prescod There's no PR description, and the test makes no assertion. What is the change in behavior that you're implementing here?

davisagli: @prescod thanks for adding a PR description but I'm still not clear: does this silently leave data out of the continuation file, or does it somehow work even though those references are not included? It would be helpful to have an assertion in this test about what data was generated.

prescod: I had also added a comment to the code. Does it help?

"""    Do not include related ObjectRows because circular
        references in serialization formats cause problems."""

The gist is that if object A refers to object B, object A will be serialized but B will not (unless it independently would have been). This prevents A->B->A loops.

I don't know of a use-case where maintaining the A->B relationship across continuations is useful. If you need access to B across continuations you can give it a nickname.

davisagli: > I don't know of a use-case where maintaining the A->B relationship across continuations is useful.

This is the piece I was looking for confirmation of.

davisagli (approved):

PR #166, opened 2020-10-30
Official embedding API

prescod:
Create an embedding API for using Snowfakery from Python apps. CumulusCI will be updated to use this API.


snowfakery/__init__.py
```diff
@@ -10,3 +11,35 @@
 version_file = Path(__file__).parent / "version.txt"
 with version_file.open() as f:
     version = f.read().strip()
+
+
+def generate_data(
+    yaml_file: Union[Path, str],
+    option: List[Tuple[str, str]] = [],
```
davisagli, line 18: Why is this name singular? And why is it a list instead of a dict?

snowfakery/__init__.py
```diff
@@ -10,3 +11,35 @@
 version_file = Path(__file__).parent / "version.txt"
 with version_file.open() as f:
     version = f.read().strip()
+
+
+def generate_data(
```
davisagli, line 16: This is a lot of args. Should we add `*` to force the caller to use named keyword arguments, so that it's not possible to use positional args which could be unclear and would make it hard for us to change the order?

snowfakery/__init__.py
```diff
@@ -10,3 +11,35 @@
...
+    debug_internals: bool = False,
+    output_format: str = None,
+    output_file: Path = None,
+    output_folder: Path = None,
+    continuation_file: Path = None,
+):
+    from .cli import generate_cli
+
+    if target_number:
+        assert isinstance(target_number[0], int)
```
davisagli, line 30: Could maybe use the `@typechecked` decorator from typeguard to do runtime checks against the type annotations? :)

snowfakery/__init__.py
```diff
@@ -10,3 +11,35 @@
...
+    yaml_file: Union[Path, str],
+    option: List[Tuple[str, str]] = [],
+    dburl: str = None,
+    target_number: Optional[Tuple[int, str]] = None,
+    debug_internals: bool = False,
+    output_format: str = None,
+    output_file: Path = None,
+    output_folder: Path = None,
+    continuation_file: Path = None,
+):
```
davisagli, line 26: The GenerateDataFromYaml task currently calls `snowfakery.data_generator.generate` with some arguments that are not included here: `stopping_criteria` and `generate_continuation_file`. How will those be handled when you update CumulusCI to use the new API?

snowfakery/__init__.py
prescod: I was translating from the click-derived API too literally. Fixed now. "user_options".

4119821

snowfakery/__init__.py
prescod: Good idea.

snowfakery/__init__.py
prescod: stopping_criteria is an internal datatype generated from the target_number. So when CCI stops using Snowfakery internals it will use target_number.

generate_continuation_file was an oversight.

I've also decided to add generate_cci_mapping_file even though only one tool will use it...

davisagli (approved):

PR #174, opened 2020-11-11
Handle some corner cases

prescod:
A couple of corner cases which changed in recent versions.

Allow "none" as a return code from random_choice because sometimes NULL is a real choice.

Allow random_choice with a single item. It previously allowed this. Machine-generated Snowfakery files may sometimes generate a single-item random_choice.


prescod:
This function has 100% coverage despite coveralls being confused.

davisagli (approved):

davisagli (changes requested):
Wait no, I'm not approving this on your holiday!

davisagli (approved):

PR #182, opened 2020-11-24
WIP: Give feedback as records are created

prescod:
Give user feedback as records are created

Next step: tests.


davisagli:
Might be fun to use https://click.palletsprojects.com/en/7.x/utils/#showing-progress-bars

PR #183, opened 2020-11-24
Fixes relating to continuation handling.

prescod:
The Snowfakery docs say:

> references are always to objects
created within the current "batch" of a recipe and never to a previous
batch with only one exception: objects marked 'just_once' are always
created only in the first run and references to them are to the
objects created in that first batch.

Given this, there is no point keeping track of objects created in previous iterations of the recipe other than those with nicknames, so Snowfakery does not do that anymore.

Also fixes two bugs which could cause crashes relating to the intersection of a) forward references, b) nicknames and c) continuations.

One solution is not to serialize "nickname slots" into the continuation file because they add no value and we are not trying to serialize deep hierarchies of ObjectRows.

A second bugfix is to ensure that if a persistent (just_once) nickname fulfills a nickname slot then it grabs the slot as soon as the interpreter is re-initialized.

Finally, made the repr() for an object more robust so that even buggy objects can print themselves out for debugging.

### DIFF
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -207,12 +207,18 @@ class Globals:
             Dependency(table_name_from, table_name_to, fieldname)
         )
 
-    def reset_slots(self):
+    def reset_slots(self, nicknamed_objects: Dict = None):
         "At the beginning of every execution, reset the forward reference slots"
         self.named_slots = {
             name: NicknameSlot(table, self.id_manager)
             for name, table in self.nicknames_and_tables.items()
         }
+        if nicknamed_objects:
+            nicknames_with_slots = set(nicknamed_objects.keys()).intersection(
+                self.named_slots.keys()
+            )
+            for name in nicknames_with_slots:
+                self.named_slots[name].consume_slot()
 
     def check_slots_filled(self):
         not_filled = [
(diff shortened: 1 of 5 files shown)


snowfakery/data_generator_runtime.py
```diff
@@ -207,12 +207,18 @@ def register_intertable_reference(
...
+    def reset_slots(self, nicknamed_objects: Dict = None):
         "At the beginning of every execution, reset the forward reference slots"
         self.named_slots = {
             name: NicknameSlot(table, self.id_manager)
             for name, table in self.nicknames_and_tables.items()
         }
+        if nicknamed_objects:
+            nicknames_with_slots = set(nicknamed_objects.keys()).intersection(
+                self.named_slots.keys()
+            )
```
davisagli, line 219: I just experimented and found that this can also be written as `{*nicknamed_objects}.intersection(self.named_slots)`

The way you have it is probably more obvious though.

davisagli (approved):
Looks reasonable. Should we document the fact that the continuation file format might change between versions of Snowfakery, or add a version marker so that we can throw an error early if you try to use an older continuation file?

prescod:
Yes I will definitely document the backwards incompatibility in the history file. Honestly the continuation file feature is barely documented and is really just there for me and CumulusCI at that this point.

PR #184, opened 2020-11-27
random_reference function

prescod:
Implements a feature for selecting a random row instead of just the last one.

The PR is largish but not as large as it seems because some code moved to another file.

There is an experimental undocumented feature for selecting a record from past iterations but it's unclear if anyone needs the feature so it outputs a warning on stderr.


atrancandoris:
This covers my use cases, which all fall under the `scope: "local"` distinction. The distribution for `scope: "global"` seems a bit odd and counter intuitive, but I am not familiar enough with the design challenges, which I assume are why it was developed this way.

Thanks for coming up with a solution!

acrosman:
I think the implementation works but the language on the definition of an "iteration" may need refinement. Although this may be a larger question in the docs and not just this section. 

Are we talking iteration of the recipe or iteration of the object?  

If I had a recipe that generates a set of objects, then later generates another set with a different pattern of details, would a random_reference call with a local scope after that second iteration reference to both sets or just the second?

``` yaml
- object: Owner
  count: 10
  fields:
    name: fake.name
    likes_pets: No

- object: Owner
  count: 10
  fields:
    name: fake.name
   likes_pets: Yes

- object: Pet
  count: 10
  fields:
    ownedBy:
        random_reference:
          tablename: Owner
          scope: local
```

Are the pet hating people above going to get pets anyway?

prescod:
> This covers my use cases, which all fall under the `scope: "local"` distinction. The distribution for `scope: "global"` seems a bit odd and counter intuitive, but I am not familiar enough with the design challenges, which I assume are why it was developed this way.

It all revolves around the fact that Snowfakery can only link to objects that were already created. If you imagine a recipe that creates two records A and B with B's linking to any random A.

On the first iteration there is only one A and one B. Therefore the chance of A being linked to is 100%

On the second iteration there are two A's and a new B. The chance of the original A being linked to again is 50%

On the there iteration there are three A's and a new B. The chance of the original A being linked to again is 30%. So it has at least 1, probably 2 and maybe 3 incoming links. Meanwhile the newest A has only a 30% chance of having an incoming link at all and a zero percent chance of having 2 or more links.

prescod:
@acrosman @atrancandoris I'm thinking that maybe I'll just delete the global mode and keep things simple. Nobody has specifically asked for two modes yet and for all I know there is no need for it in most use-cases.

This does imply that you cannot use it with extremely simple recipes like this:

```
- object: Account
  .... 
- object Contact
  fields:
     AccountId:
         random_reference: Account
```

Because there's only one Account to link to in every iteration of the recipe.

Instead you'd need:

```
- object: Account
   contact: 50
  .... 
- object Contact
   contact: 50
  fields:
     AccountId:
         random_reference: Account
```

Now your connections will "seem" random.

acrosman:
If I understand where this is, it is going to allow random references to objects generated previously within the recipe.  If the recipe is being run multiple times each iteration only references other objects in that iteration.  

```yaml
- object: Person
  count: 10
  fields:
    name: fake.name
    likes_pets: Yes

- object: Person
  count: 10
  fields:
    name: fake.name
    likes_pets: No

- object: Pet
  count: 10
  fields:
    ownedBy:
        random_reference: Person
```

In this case the pets would be spread across all 20 people without consideration for their preference.

But if we shuffle order of generation: 
```yaml
- object: Person
  count: 10
  fields:
    name: fake.name
    likes_pets: Yes

- object: Pet
  count: 10
  fields:
    ownedBy:
        random_reference: Person

- object: Person
  count: 10
  fields:
    name: fake.name
    likes_pets: No
```
the first 10 people have a pet and the second 10 do not.  Which allows people to have some control of similar objects and differentiating which get relationships for at least some use cases.

tests/test_random_choice_calculated.yml
```diff
@@ -0,0 +1,8 @@
+- object : A
+  fields:
+    __a_percent: 60
+    b:
+        random_choice:
+            A: ${{__a_percent}}%
+            B: ${{__a_percent / 2}}%
+            C: ${{100 - (__a_percent * 1.5)}}%
```
davisagli, line 8: This file isn't referenced anywhere. Accidentally added from other work?

tests/test_references.py
```diff
@@ -205,3 +204,151 @@ def test_forward_references_not_fulfilled__nickname(self, write_row):
...
+        yaml = """                  #1
+      - object: A                   #2
+        count: 10                   #4
+      - object: B                   #5
+        count: 2                    #6
+        fields:                     #7
+            A_ref:                  #8
+              random_reference:
+                tablename: A   #9
+                scope: global
```
davisagli, line 345: `scope: current-iteration` (default) vs `scope: prior-and-current-iterations` might be clearer naming? (if a little verbose)

tests/conftest.py
```diff
@@ -12,8 +12,25 @@ def generated_rows(request):
     def row_values(index, value):
         return mockobj.mock_calls[index][1][1][value]
 
+    def table_values(tablename, index, field=None):
+        "Look up a value from a table."
```
davisagli, line 16:
```suggestion
        """Look up a value from a table."""
```

(I know it's equivalent, but the visual consistency in how docstrings appear is helpful when scanning through code.)

tests/conftest.py
```diff
@@ -12,8 +12,25 @@ def generated_rows(request):
...
+        index = index - 1  # use 1-based indexing like Snowfakery does
+
+        # create and cache a dict of table names to lists of rows
+        if type(mockobj._index) != {}:
```
davisagli, line 20: Surely you mean

```suggestion
        if type(mockobj._index) != dict:
```

snowfakery/data_generator_runtime.py
```diff
@@ -108,52 +108,17 @@ class Dependency(NamedTuple):
...
+    def __init__(self, nicknames_and_tables: Mapping[str, str], id_manager: IdManager):
+        self.nicknamed_objects = {}
+        self.last_seen_obj_by_table = {}
+        self.named_slots = {
+            name: NicknameSlot(table, id_manager)
+            for name, table in nicknames_and_tables.items()
+        }
+        self.orig_used_ids = defaultdict(
+            lambda: 1, {k: v + 1 for k, v in id_manager.last_used_ids.items()}
+        )
```
davisagli, line 121: This might warrant a comment -- took me a bit to figure out how this is being used.

Would it be worth changing the id manager to use 1-based indexing too?

snowfakery/data_generator_runtime.py
```diff
@@ -234,37 +212,57 @@ def check_slots_filled(self):
...
+        def serialize_dict_of_object_rows(dct):
+            return {k: v.__getstate__() for k, v in dct.items()}
+
+        persistent_nicknames = serialize_dict_of_object_rows(self.persistent_nicknames)
+        persistent_objects_by_table = serialize_dict_of_object_rows(
+            self.persistent_objects_by_table
+        )
         intertable_dependencies = [
             dict(v._asdict()) for v in self.intertable_dependencies
-        ]  # converts ordered-dict to dict for Python 3.6 and 3.7
```
davisagli, line 246: This comment is still helpful.

snowfakery/data_generator_runtime.py
```diff
@@ -360,12 +357,17 @@ def __init__(
 
         self.templates = templates
 
-    def loop_over_templates_until_finished(self, runtimecontext, continuing):
+    # TODO: move this into the interpreter object
+    def loop_over_templates_until_finished(self, continuing):
         finished = False
+        # TODO: make a fresh runtime context per execution to avoid
+        #       risk of data leakage
+        runtimecontext = RuntimeContext(interpreter=self)
```
davisagli, line 365: Looks like these todos were done?

snowfakery/data_generator_runtime.py
prescod: Why don't you explain it to me so I can comment it. :) (just kidding....)

tests/test_random_choice_calculated.yml
prescod: Removed in f3b2629

tests/test_references.py
prescod: Renamed in f3b2629 . (It's still a secret feature so I'm not going to workshop the name too much but your suggestion is clear and could always be changed/aliased later)

tests/conftest.py
prescod: Fixed in f3b2629

snowfakery/data_generator_runtime.py
prescod: I think its much clearer in f3b2629 and the new naming (last_used_id versus first_new_id) probably clarifies why one should be 0-based and the other 1-based.

snowfakery/data_generator_runtime.py
prescod: Restored in f3b2629.

snowfakery/data_generator_runtime.py
prescod: Removed in 0730497

snowfakery/data_generator_runtime.py
davisagli: @prescod yes, thanks, this is much clearer.

snowfakery/data_generator_runtime.py
davisagli: @prescod I meant "# TODO: move this into the interpreter object" too

prescod: Makes sense. 0234afd

davisagli (approved):

PR #185, opened 2020-11-28
Feature/use native types

prescod:
The way Jinja-in-Snowfakery was configured before, all objects were serialized to strings before being captured as fields. This change defers the serialization to the output_stream which allows operations like:

foo: ${{OtherObj.IntField + 5}}

or

foo: ${{OtherObj.DateField + relativedelta(days=3}}

etc.

There is also a subtle but important benefit to allowing Snowfakery to detect the "real" type of:

foo: ${{OtherObj.OtherObj}}

I also added a `debug` function while I was debugging the Python 3.6 issue.


prescod:
Other branches will increase template_funcs to 100% coverage and hopefully `coverage` will stop complaining about it.

prescod:
Note to self: test NativeTypes with StringGenerators. NativeTypes may disable the serialization which triggers StringGenerators and thus might break plugins or types that depend on it.

PR #186, opened 2020-11-28
Allow dotted references (without Jinja)

prescod:
Previously snowfakery "references" had to be either simple strings or Jinja expressions. This allows a simpler dotted syntax to eliminate the need for Jinja in many cases. Future versions of Snowfakery may also be able to do a bit more static analysis of dotted-references than of Jinja expressions.


docs/index.md
```diff
@@ -418,6 +418,30 @@ nickname or tablename, that object is the target reference.
 Otherwise, the reference can be to an object that has not been created yet. Snowfakery will generate an ID for the object so that the current row can be generated. No other properties of the other object can be referred to, because
 it does not exist yet.
 
+If the referenced name contains dots then Snowfakery will follow
+field names to get to the final name. The first name before the first "." could be either a field name or an object-name.
```
davisagli, line 422: Are we okay with changing behavior if someone was already using dots in names?

snowfakery/template_funcs.py
```diff
@@ -130,14 +130,26 @@ def reference(self, x: Union[ObjectRow, str]):
...
             elif isinstance(x, str):  # name of an object
-                obj = self.context.field_vars().get(x)
-                if not obj:
+                # allows dotted paths
+                parts = x.split(".")
+                target = self.context.field_vars().get(parts.pop(0))
+
+                for part in parts:
+                    try:
+                        target = getattr(target, part)
```
davisagli, line 139: We should make sure it's not possible to traverse to dunder attributes this way.

snowfakery/template_funcs.py
```diff
@@ -130,14 +130,26 @@ def reference(self, x: Union[ObjectRow, str]):
...
+                    except AttributeError:
+                        raise DataGenError(
+                            f"Expression cannot be evaluated `{x}``. Problem before `{part}`: {target}",
+                            None,
+                            None,
```
davisagli, line 144: Should DataGenError default to None for the 2nd and 3rd args so it's not necessary to spell them out?

docs/index.md
prescod: Snowfakery did always use a Python-meaning for dots in Jinja. Also I don't know of any users using Snowfakery outside of the SalesForce universe. But I'll note it as a potential backwards incompatibility.

snowfakery/template_funcs.py
prescod: It probably is but it always was in Jinja. I don't think we should think of Snowfakery files as being any more secure than Python code (at least as long as we run them in a Python runtime wth Jinja enabled).

Traversing through leadiing-double-underscore is probably useful because Snowfakery uses them but I could disallow traversing through trailing-double-underscore identifiers.

Using Jinja:

```Bar(id=1, bar=<bound method ObjectRow.__getattr__ of <ObjectRow Foo 1>>)```

I should probably add a note to the docs about only executing trusted recipes or recipes in a hardened context like a container.

snowfakery/template_funcs.py
prescod: Maybe. I wanted to reduce the likelihood of forgetting to supply them when they were available but the deeper you get into the interpreter, the less likely it is that they are available.

snowfakery/template_funcs.py
davisagli: For jinja2 we can and should be using the sandboxed environment: https://jinja.palletsprojects.com/en/2.11.x/sandbox/

And if we don't do that, then yes, we should make it a lot clearer that running a snowfakery recipe is equivalent to running code.

snowfakery/template_funcs.py
davisagli: Isn't it the sort of thing where the outer layers of the interpreter should keep track of the context and be responsible for annotating any exceptions raised by the inner layers?

Making them required hasn't solved the problem of forgetting them: https://github.com/SFDO-Tooling/Snowfakery/pull/161#discussion_r508553754

snowfakery/template_funcs.py
prescod: Unfortunately it doesn't seem to be possible to simultaneously get the benefits of both the NativeEnvironment and the SandboxEnvironment and I have a PR open to use the NativeEnvironment.

#185

snowfakery/template_funcs.py
prescod: #273

snowfakery/template_funcs.py
davisagli: Okay. Waiting for you to add that clear warning to the docs then.

prescod: 8a14ebf37f26d24eea3eb429ba285c6fe0d99561

davisagli (approved):

PR #188, opened 2020-11-29
Add int, min, round to the math plugin

prescod:
Add int, min, round to the math plugin


tests/test_template_funcs.py
```diff
@@ -253,6 +253,19 @@ def test_functions_inline(self, write_row):
...
+        yaml = """
+        - object : A
+          fields:
+            date:
+                date_between:
+                    start_date: "2040-13-13"
+                    end_date: "2040-13-13"
+        """
+        with pytest.raises(DataGenError):
+            generate(StringIO(yaml), {}, None)
```
davisagli, line 267: I don't understand how this test relates to the rest of the diff

prescod: It was included by accident while I was investigating another issue but its a useful test (better name now) so I think I should leave it.

davisagli (approved):

PR #191, opened 2020-11-30
Update HISTORY.md

prescod:
I noticed a typo after release and while I was fixing it I did some reordering of the release notes. PyPi doesn't do anything with this history file so editing after release shouldn't have negative consequences.


davisagli (approved):
Heh, I actually noticed that typo at one point yesterday, and then couldn't find it when I returned to do the review.

PR #213, opened 2021-01-15
Fix link

prescod:
If you can't test a link you should be extra-careful in typing it in the first place!


davisagli (approved):

davisagli (commented):
Don't forget to update this when you switch the default branch to main!

PR #218, opened 2021-01-18
Allow fields to be specified in any order

### LINKED ISSUE #205: Using fields from a macro in a formula.
When trying to use fields on an object that come from a macro in a forumla, the formula evaluates as blank instead of with the expected field value.  I'm doing this in a nested macro context so that may be part of the challenge (and makes the example below a hair more complicated)

Example:  
If I have a Macro that create name fields:
```
- macro: sfContact
  fields:
    Salutation:
      fake: prefix
    FirstName:
      fake: first_name
    LastName:
      fake: last_name
...
```

Trigger that within another macro

```
- macro: HedStudent
  include: sfContact
  fields:
    Title: None
    hed__UniversityEmail__c: ${{FirstName}}.${{LastName}}@example.edu
...
```

And finally create my object:  
```
- object: Contact
  include: HedStudent
...
```

I expected `[FirstName].[LastName]@example.edu`
I got `.@example.edu`

prescod:
Snowfakery fields can now be specified in any order and formulas will still find them.

This is implemented through two mechanisms:

1. Fields are sorted before evaluation based on their relationships to each other (handles most cases)
2. Fields have lazy proxies (lambdas) that stand in for them in case their values are needed before they are evaluated (handles some corner cases relating to nested objects)

This version should be performance-tested because the lazy proxies have a cost.

Fixes #205

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ PyYAML==5.3.1
 click==7.1.2
 python-dateutil==2.8.1
 gvgen==1.0
+networkx==2.5
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -510,6 +510,9 @@ class RuntimeContext:
     def context_vars(self, plugin_namespace):
         return self._context_vars.setdefault(plugin_namespace, {})
 
+    def object_names(self):
+        return self.evaluation_namespace.object_names()
+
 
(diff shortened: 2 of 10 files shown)


prescod:
The performance cost does seem quite high so I'll need to do some optimization.

prescod:
This feature has not been asked for urgently and we will not move on with it at this time.

PR #229, opened 2021-01-25
Search for plugins in a few places

prescod:
Snowfakery will now look for a declared plugin or provider in these places:

- the [Python path](https://docs.python.org/3/library/sys.html#sys.path),
- in a `plugins` directory in the same directory as the recipe,
- in a `plugins` directory below the current working directory and
- in a subdirectory of the user's home directory called `.snowfakery/plugins`.


prescod:
@acrosman . Does this make sense to you?

acrosman:
Yes it does.  Only other thing I might consider would be an command line option to add something for the person who wants some other option.  But this certainly addresses the need I ran into.

prescod:
Yes I'm sure we'll want that at some point but this is probably enough for the next six months or so.

snowfakery/plugins.py
```diff
@@ -102,6 +111,23 @@ def lazy(func: Any) -> Callable:
...
+    cwd_plugins = "./plugins"
+    user_plugins = Path.home() / ".snowfakery/plugins"
+    new_sys_path = [
+        *sys.path,
+        *(str(p) for p in search_paths),
+        str(cwd_plugins),
+        str(user_plugins),
+    ]
+
+    with patch.object(sys, "path", new_sys_path):
```
davisagli, line 127: This looks potentially dangerous if snowfakery is ever embedded in a threaded context, but I'm not sure what to suggest as an alternative.

davisagli (approved):

PR #231, opened 2021-01-26
Add variables to Snowfakery

prescod:
Snowfakery now has a notion of variables.

docs/index.md describes the feature.

Also: my internal variable scoping feature has been replaced with a version based on Python's ChainMap.

This is the first feature in Snowfakery that allows two different ordered constructs to interleave and affect each other so a new concept of Statement is introduced. Other statements may be added over time.

### DIFF
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -357,7 +357,7 @@ class Interpreter:
 
     def loop_over_templates_until_finished(self, continuing):
         finished = False
-        runtimecontext = RuntimeContext(interpreter=self)
+        RuntimeContext(interpreter=self)
         while not finished:
             self.loop_over_templates_once(self.statements, continuing)
             finished = self.current_context.check_if_finished()
@@ -366,18 +366,11 @@ class Interpreter:
 
     def loop_over_templates_once(self, statement_list, continuing: bool):
         for statement in statement_list:
-            if statement.__class__.__name__ == "ObjectTemplate":
-                template = statement
-                should_skip = template.just_once and continuing
-                if not should_skip:
-                    template.generate_rows(self.output_stream, self.current_context)
-            elif statement.__class__.__name__ == "VariableDefinition":
-                vardef = statement
-                self.current_context.variable_definitions()[
-                    vardef.varname
-                ] = vardef.evaluate(self.current_context)
-            else:
-                assert 0, statement  # pragma: no cover
+            statement.execute(self, self.current_context, continuing)
+
+    def register_variable(self, name: str, value: object):
+        vardefs = self.current_context.variable_definitions()
+        vardefs[name] = value
 
     def __enter__(self, *args):
         return self
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -357,7 +357,7 @@ class Interpreter:
 
     def loop_over_templates_until_finished(self, continuing):
         finished = False
-        RuntimeContext(interpreter=self)
+        self.current_context = RuntimeContext(interpreter=self)
         while not finished:
             self.loop_over_templates_once(self.statements, continuing)
             finished = self.current_context.check_if_finished()
@@ -458,9 +458,9 @@ class RuntimeContext:
             interpreter=self.interpreter,
             parent_context=self,
         )
-        # jr will register itself with the interpreter
         try:
             yield jr
+            self.interpreter.current_context = jr
         finally:
             # Goodbye junior, its been nice
             # Hope you find your paradise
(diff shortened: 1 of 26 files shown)


snowfakery/data_generator_runtime.py
```diff
@@ -356,22 +346,38 @@ def __init__(
...
+    def faker_template_library(self, locale):
+        rc = self.faker_template_libraries.get(locale)
+        if not rc:
+            rc = FakerTemplateLibrary(self.faker_providers, locale)
+            self.faker_template_libraries[locale] = rc
+        return rc
 
     def loop_over_templates_until_finished(self, continuing):
         finished = False
         runtimecontext = RuntimeContext(interpreter=self)
```
davisagli, line 360: It was not obvious to me that constructing a RuntimeContext updates the interpreter's `current_context`. Perhaps that could be made explicit?
```suggestion
        self.current_context = RuntimeContext(interpreter=self)
```

(And remove the implicit setting within RuntimeContext's `__init__`)

snowfakery/data_generator_runtime.py
```diff
@@ -356,22 +346,38 @@ def __init__(
...
+                should_skip = template.just_once and continuing
+                if not should_skip:
+                    template.generate_rows(self.output_stream, self.current_context)
+            elif statement.__class__.__name__ == "VariableDefinition":
+                vardef = statement
+                self.current_context.variable_definitions()[
+                    vardef.varname
+                ] = vardef.evaluate(self.current_context)
+            else:
+                assert 0, statement  # pragma: no cover
```
davisagli, line 380: The if/else checking class types makes me think this should be implemented as an `evaluate` method of the statement object (passed the interpreter as an arg, I suppose).

snowfakery/parse_recipe_yaml.py
```diff
@@ -315,16 +325,46 @@ def parse_object_template(yaml_sobj: Dict, context: ParseContext) -> ObjectTempl
...
+    yaml_sobj: Dict, context: ParseContext
+) -> VariableDefinition:
+    parsed_template: Any = parse_element(
+        yaml_sobj,
+        "var",
+        {},
+        {
+            "value": (str, int, dict, list),
+        },
+        context,
```
davisagli, line 338: Please use keyword arguments here.

snowfakery/parse_recipe_yaml.py
```diff
@@ -315,16 +325,46 @@ def parse_object_template(yaml_sobj: Dict, context: ParseContext) -> ObjectTempl
...
+    )
+
+    assert yaml_sobj
+    with context.change_current_parent_object(yaml_sobj):
+        sobj_def = {}
+        sobj_def["varname"] = parsed_template.var
```
davisagli, line 344: Hmm, it's typed as `Any`, but we obviously expect it to have a `var`. Doesn't need to be in this PR, but perhaps we can replace `parse_element` with something that is typed more specifically.

snowfakery/template_funcs.py
```diff
@@ -68,6 +66,7 @@ def render_boolean(context: PluginContext, value: FieldDefinition) -> bool:
 class StandardFuncs(SnowfakeryPlugin):
     class Functions:
         int = int
+        _faker_for_dates = Faker(use_weighting=False)  # use ONLY for random_dates
```
davisagli, line 69: This makes me wonder _why_ I shouldn't use it for something else

snowfakery/parse_recipe_yaml.py
prescod: parse_element is the boundary between the untyped YAML world and the dynamically typed world. It's more or less my "poor man's Pydantic". So its return type will always be polymorphic.

snowfakery/template_funcs.py
prescod: 5e9be1e1a2bafbf654f2ff3d441727363f540c5a

snowfakery/data_generator_runtime.py
```diff
@@ -356,22 +346,38 @@ def __init__(
...
 
-        self.templates = templates
+        self.statements = statements
+
+    def faker_template_library(self, locale):
+        rc = self.faker_template_libraries.get(locale)
+        if not rc:
+            rc = FakerTemplateLibrary(self.faker_providers, locale)
+            self.faker_template_libraries[locale] = rc
+        return rc
```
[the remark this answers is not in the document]
davisagli: Oh yeah, I guess I made that up.

snowfakery/parse_recipe_yaml.py
davisagli: @prescod ok, I get that parse_element needs to return different types. For a case like this where a particular type is _expected_, it would still be possible to add a type hint on the variable that the return value is being assigned to (along with perhaps an assertion to confirm that it's the expected type at runtime).

Anyway, not clear that's worth the effort here. I'll resolve this.

davisagli (approved):

PR #243, opened 2021-02-01
Docs about using 3rd party databases

prescod:


docs/index.md
```diff
@@ -1500,6 +1500,64 @@ $ cci task run generate_opportunities_and_contacts
...
+
+Snowfakery is built on top of a very flexible engine called
+SQLAlchemy. This allows it to connect to many different databases
+subject to the limitations described below.
+
+You should start by installing Snowfakery in a context which
+makes it easy to use the Python command 'pip' to manage your
+Python environment. For example you could install Python
+using the standard installers from `python.org` and then
+you would do these commands to create and use a venv with the
```
Br4nd0R, line 1513:
```suggestion
you would run the following commands to create and use a venv with the
```

docs/index.md
```diff
@@ -1500,6 +1500,64 @@ $ cci task run generate_opportunities_and_contacts
...
+$ echo 'select * from "Employee"' | psql snowfakerydb
+```
+
+That's a lot to take in, but hopefully it will be clear enough
+to follow the links and understand the details.
+
+A limitation of this process is that currently Snowfakery can
+only create new tables rather than import into existing ones.
+
+The table will have an id column and columns for every field that
```
Br4nd0R, line 1547: Wording is slightly confusing at first. Maybe
```suggestion
The table will have an id column, in addition to columns for every field that
```

Br4nd0R (approved):
A couple minor wording suggestions. I'll go ahead and approve though, looks good!

Br4nd0R (approved):

PR #247, opened 2021-02-03
Snowfakery 1.5 and locale docs/examples

prescod:
I realized that the locale feature had slipped in without proper docs and examples so I added them below.

I also changed the variable name from 'locale" to 'snowfakery_locale' on the premise that it is easier to learn the language if you can easily distinguish built-ins from user-vars.


docs/index.md
```diff
@@ -557,28 +557,28 @@ You can also call these functions with arguments as described in Faker's [docume
...
+You can specify internationally appropriate fakes for many different kind of names (e.g. person, company) by setting the snowfakery_locale this:
 
 ```yaml
-- object: Viking
+- var: snowfakery_locale
+  value: no_NO
+- object: person
   fields:
-    Name:
-      i18n_fake:
```
davisagli, line 568: @prescod Is this syntax no longer supported? If so that should probably be called out in the release notes.

snowfakery/template_funcs.py
```diff
@@ -121,6 +121,7 @@ def try_parse_date(d):
             # swallow empty range errors per Python conventions
 
         def i18n_fake(self, locale: str, fake: str):
+            # deprecated by still here for backwards compatibility
```
davisagli, line 124: Ah, this answers my question above. Is there still a test that uses it so we are sure not to accidentally break it?

davisagli: coverage says yes.

davisagli (approved):

PR #249, opened 2021-02-04
Fix docs about context bugs and add template_filename variable

prescod:
The documentation referred to variables that don't actually exist. So I made some minor naming changes, fixed the docs and added tests to ensure that the documented features don't break.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -727,9 +727,7 @@ a single recipe.
 
 The `fake` variable gives access to faker as described elsewhere in this documentation.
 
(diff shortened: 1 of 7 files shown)


davisagli:
@prescod the tests failed

prescod:
Okay I renamed it again and also ensured that you can get the filename in a 'var definition' for consistency. Now it is `snowfakery_filename` and you can fetch it in a variable definition or an object template.

davisagli (approved):

Issue #239, opened 2021-01-30
Add "Null" to templates and docs


prescod:
YAML uses NULL. Inject it into formulas and document how to use it in Snowfakery.

PR #250, opened 2021-02-04
Add NULL to Snowfakery formulas #239

prescod:
YAML uses NULL so now Snowfakery also use NULL in Jinja templates.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -745,8 +745,70 @@ from `dateutil` is available for use in calculations like this:
 ${{ date(Date_Established__c) + relativedelta(months=child_index) }}
 ```
 
-Some plugins may also be interested in a `template` variable which has an `id` attributes represents a unique identifier for the current template. Look at 
+Some plugins may also be interested in a `template` variable which has an `id` attributes represents a unique identifier for the current template. Look at
 [datasets.py](https://github.com/SFDO-Tooling/Snowfakery/blob/master/snowfakery/standard_plugins/datasets.py) to see one use-case where the template's ID can used to differentiate between two otherwise identical datasets.
+
+### NULL
+
+The value `NULL` can be used to represent a missing value as in the recipe
+below:
+
+First, we can make `NULL` an option with `random_choice`:
+
+```yaml
+- object: foo
+  count: 10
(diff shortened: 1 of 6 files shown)


davisagli (approved):

docs/index.md
```diff
@@ -745,9 +745,71 @@ from `dateutil` is available for use in calculations like this:
...
+    DateSupplied:
+      if:
+        - choice:
+            when: ${{ EndDate!=NULL }}
+            pick: "Yes"
+        - choice:
+            pick: "No"
+```
+
+### NULL
```
mbeaty-sf, line 782: Looks like the block has been duplicated from line 782 to 811

davisagli (changes requested):
@prescod waiting for you to resolve the issue Michael pointed out

prescod:
@davisagli Thanks for the reminder. I forgot to push that change.

tests/test_references.py
```diff
@@ -419,3 +419,16 @@ def test_reference_wrong_type(self):
...
+    def test_reference_really_wrong_type(self, disable_typeguard):
+        yaml = """
+        - object: B
+          count: 2
+          fields:
+              A_ref:
+                reference: 5"""
+        with mock.patch(
+            "typeguard.check_argument_types", lambda *args, **kwargs: ...
+        ), pytest.raises(DataGenError) as e:
```
davisagli, line 432: It looks to me like it would actually be appropriate to type the `reference` function as accepting Any, because it does have logic for handling any type.

Also you're patching typeguard twice

davisagli (changes requested):

tests/test_references.py
prescod: Okay, I switched to Any to make it easier to switch to MyPy some day.

prescod:
I think I will use the disable_typeguard function in the future though. I've run into similar problems in the past.

davisagli:
> I think I will use the disable_typeguard function in the future though. I've run into similar problems in the past.

Please only if we conclude that it's impossible to express the types accurately.

davisagli (approved):

PR #257, opened 2021-02-09
Tool for merging example files into docs

prescod:
This tool merges example files into the docs and thus ensures that the tested version of the example is also the one in the doc.

### DIFF
--- a/tools/docs_samples.py
+++ b/tools/docs_samples.py
@@ -1,0 +1,89 @@
+from pathlib import Path
+
+# This tool merges example files into the docs and thus ensures
+# that the tested version of the example is also the one in the
+# doc.
+
+
+class TOP:
+    """Top level of the Markdown doc"""
+
+    def next_state(line, location):
+        if line.startswith(TRIPLE_QUOTE):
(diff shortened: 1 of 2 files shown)


tools/docs_samples.py
```diff
@@ -0,0 +1,89 @@
...
+            with Path(filename).open() as f:
+                example_lines = f.readlines()
+            print("Replacing", filename)
+            return REPLACING, [line] + example_lines
+        else:
+            return OPEN, [line]
+
+
+def check_triple_quote_alone(line, location):
+    """Triple-quotes at the ends or mmiddle of line are
```
Br4nd0R, line 56:
```suggestion
    """Triple-quotes at the ends or middle of line are
```

Br4nd0R (dismissed):
LGTM!

Br4nd0R (approved):

davisagli (commented):
I appreciate the state machine, but would it work to use the `literalinclude` directive for this use case? https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

PR #270, opened 2021-02-26
Simpler record type handling

prescod:
I realized that there is a much simpler way to do record types. I guess the first time I implemented it, I didn't really understand CCI's expected format. Now I just fake that format.

CCI will also need a small change to be compatible.


davisagli (approved):
This looks like a much better approach.

PR #288, opened 2021-03-17
Dependencies updates.

prescod:
Faker is the only updated dependency that is relevant to end-users, and it is a minor change.


codexmarcel (approved):
LGTM

PR #290, opened 2021-03-17
Mapping file generation super-powers

prescod:
1. Allow users to inject load declarations into generated mapping file.
2. Allow users to order mapping file declarations
3. Auto-inject "after:" declarations where appropriate.

(the description below is the vision...not every detail is implemented yet)

Still needs CCI side, documentation and tests.

### Snowfakery and Mapping Files

CumulusCI's bulk loader depends on a mapping file to determine the load order of data.

Snowfakery generates this mapping file automatically in the `generate_and_load_from_yaml` job. Occasionally you may need to inject some additional information into the load. For example, you might want to change which API or processing mode to use (REST API, BULK API, Bulk Serial, Bulk Parallel). Or you might want to enforce a certain load order.

You can now do that by specifying a "recipe.load.yml" file to go with your recipe file. The --load-yml command line argument allows that and there is also a file naming convention which does it automatically (TBD).

Here are the declarations you may put into the "recipe.load.yml".

- sf_object: Contact 
  bulk_mode: serial  # same as specifying `bulk_mode` in mapping file

- sf_object: Contact
  load_after: Opportunity  # load contacts after opportunities for some reasoon

- sf_object: Account  # same as specifying `api` in mapping file
  api: rest


snowfakery/cci_mapping_files/declaration_parser.py
```diff
@@ -0,0 +1,118 @@
...
+        if existing_decl:
+            existing_decl.value.append(new_decl.value)
+            return existing_decl
+        else:
+            d = new_decl
+            # start to build a list-based declaration
+            return AtomicDecl(d.sf_object, d.key, [d.value], d.priority, d.merge_rule)
+
+
+class AtomicDecl(T.NamedTuple):
```
Br4nd0R, line 29: You can move this above `MergeRules` and add a doc string :)

snowfakery/cci_mapping_files/declaration_parser.py
```diff
@@ -0,0 +1,118 @@
...
+    "anchor_date": MergeRules.use_highest_priority,
+    "load_after": MergeRules.append,
+}
+
+
+class SObjectRuleDeclarationFile(BaseModel):
+    __root__: T.List[SObjectRuleDeclaration]
+
+    @classmethod
+    def parse_from_yaml(cls, data: T.List):
```
Br4nd0R, line 81: change name/semantics of this function

snowfakery/cli.py
```diff
@@ -131,6 +135,11 @@ def int_string_tuple(ctx, param, value=None):
     type=click.File("r"),
     help="Continue generating a dataset where 'continuation-file' left off",
 )
+@click.option(
```
Br4nd0R, line 138: Load more than one declaration file?

prescod:
> I think the feature makes sense, but why a separate file instead of including these hints somewhere in the main recipe that the mapping generator knows to look for? Do you anticipate a need to use different sets of hints with the same recipe?

Yes, I do anticipate that e.g. the standard NPSP recipe might work on your org but automation you've installed might make the load order unique to your org.

Or conversely (and perhaps more commonly!) the standard NPSP load file might work but you want to add a custom field to the recipe and therefore create your own.

In the not too distant future I could also imagine you sharing the files between Snowfakery data loads and CCI SQL loads instead of using the succinct Snowfakery format in one place and the verbose and unwieldy mapping file format in the other.

And then in the distant future, perhaps there is some centralized management of them analogous to TypeShed, DefinitelyTyped, SchemaStore.org etc. (or maybe that won't be useful enough)

snowfakery/cli.py
prescod: 12e7b03

snowfakery/cci_mapping_files/declaration_parser.py
prescod: Changed inputs.

snowfakery/cci_mapping_files/declaration_parser.py
prescod: Done and done.

snowfakery/cli.py
```diff
@@ -344,6 +352,19 @@ def stopping_criteria_from_target_number(target_number):
     return None
 
 
+from pysnooper import snoop
+
+
+@snoop()
```
Br4nd0R: TIL: `PySnooper`!

Br4nd0R (approved):
I took a look at the changes since we reviewed this together. All looks good. I really appreciated the addition of the docstrings!

davisagli:
Still needs docs, right?

prescod:
I'm planning to put the docs in the CCI repo.

PR #301, opened 2021-03-24
Fix cases where some records have a record types and others don't

prescod:
Fix a bug that is triggered when some records have a record types and others don't.


davisagli (approved):

davisagli (changes requested):
Tests are failing

davisagli (approved):

PR #305, opened 2021-03-31
Fix Github Actions

prescod:


davisagli (changes requested):
The Salesforce standard for open source repositories doesn't allow using personal access tokens for unattended access to GitHub. The alternative is probably setting up a GitHub app for this purpose (and will need a security assessment). We need to work on that for CumulusCI as well.

prescod:
@davisagli The GITHUB_TOKEN is not a personal access tokens.

"The token's permissions are limited to the repository that contains your workflow. 

Before each job begins, GitHub fetches an installation access token for the job. The token expires when the job is finished."

Did Coveralls change recently? Do you know why this problem has arisen?

davisagli:
@prescod sorry, I forgot that this is how the default GITHUB_TOKEN provided by Github Actions is accessed, so I was interpreting this as fetching a manually configured secret, and assuming that it must be a personal access token.

What is the problem you're trying to solve?

prescod:
Replacing (renewing?) the COVERALLS_REPO_TOKEN seems (?) to have fixed this problem without GITHUB_TOKEN. I'm still confused but I have some passing builds.

PR #306, opened 2021-03-31
Dependabot Roll-Up

prescod:


requirements_dev.txt
```diff
@@ -2,12 +2,12 @@
...
-flake8==3.8.4
+flake8==3.9.0
 myst-parser==0.13.5
 pre-commit==2.11.1
 pytest==6.2.2
 pytest-cov==2.11.1
-Sphinx==3.5.2
-typeguard==2.10.0
+Sphinx==3.5.3
+typeguard==2.11.0
```
davisagli, line 11: Note that cumulusci currently has typeguard held back to 2.10.0

requirements.txt
```diff
@@ -1,8 +1,8 @@
-SQLAlchemy==1.3.23
-Faker==6.6.0
+SQLAlchemy==1.4.3
```
davisagli, line 1: We can't upgrade to sqlalchemy 1.4 yet, it has more significant changes that we need to take a look at.

davisagli (commented):
Are these comments handled when these files are used as the basis for install_requires in setup.py? https://github.com/SFDO-Tooling/Snowfakery/blob/main/setup.py#L7

prescod:
Good question.

Experimentally it "seems to work" (python setup.py install) but I don't know why. I'll tweak setup.py to be careful in case it fails in some other context.

6650f3c

davisagli (approved):

PR #310, opened 2021-03-31
Snowfakery 1.8

prescod:
## Snowfakery 1.8

Fix regression when an SObject declares a RecordType on some records
but not others.  #301

Add a new feature for integrating with CumulusCI's Load architecture.
This feature will obsolete most (hopefully all) uses of mapping.yml with Snowfakery.  Documentation for it will be in CumulusCI. #209


davisagli (dismissed):

prescod:
Transposed two characters in the docs. :(

davisagli (approved):

PR #312, opened 2021-03-31
More expansive package-searching algorithm and Snowfakery 8.1

prescod:


HISTORY.md
```diff
@@ -4,7 +4,7 @@ In the beginning, programmers created the databases. Now the databases were form
 
 And so [Salesforce.org](http://salesforce.org/) said “Let there be data,” and there was Snowfakery. And it was good.
 
-## Snowfakery 1.8
+## Snowfakery 1.8 and 1.8.1
```
davisagli, line 7: I know I wouldn't appreciate a project that does this. Add a new heading with a release note like "fixed packaging issue" or something like that.

setup.py
```diff
@@ -17,6 +17,14 @@
 with open("snowfakery/version.txt") as f:
     version = f.read().strip()
 
+packages = [
+    p
+    for p in setuptools.find_namespace_packages()
+    if p.startswith("snowfakery") and not p.startswith("snowfakery.docs")
+]
+
+print("Releasing packages", packages)
```
davisagli, line 26:
```suggestion
```

Seems like an anti-pattern to use print here, and we certainly can't assume that someone is releasing something just because they're running setup.py

davisagli (changes requested):

davisagli (approved):

PR #324, opened 2021-04-10
Pre-commit hook to keep docs and examples in sync

prescod:


.pre-commit-config.yaml
```diff
@@ -10,3 +10,11 @@ repos:
...
     - id: check-merge-conflict
     - id: flake8
+- repo: local
+  hooks:
+    - id: merge-docs
+      name: Merge examples into docs
+      entry: tools/docs_samples.py
+      language: python
+      files: ".*.md"
+      always_run: true
```
davisagli, line 20: Is it hard to make it only run when there are changes in `docs` or `examples`?

prescod: 9e0864e

davisagli (approved):

PR #325, opened 2021-04-10
More reasonable nickname rules and nicer output

prescod:
Nicknames go back to allowing spaces, apostrophes and most other characters other than '.' and '"'


snowfakery/parse_recipe_yaml.py
```diff
@@ -287,6 +288,22 @@ def parse_inclusions(
         friends.extend(include_friends)
 
 
+# TODO: Use application object to write this instead
+def check_identifier(name: T.Optional[str], source: dict, context: str):
+    if not name:
+        return
+    badchars = [char for char in '."' if char in name]
+    if set(name).intersection(badchars):
```
davisagli, line 296:
```suggestion
    badchars = set('."')
    if set(name).intersection(badchars):
```

davisagli (approved):

PR #351, opened 2021-05-04
1. Fix image output. 2. Allow 3rd party outputstreams.

prescod:
Changes:

1. There was a regression with respect to the Image output formats. This fixes them.
2. Added a unit test to ensure we don't regress.
3. While I was at it, I also added the feature of pluggable output streams.


Br4nd0R (approved):
LGTM!

PR #353, opened 2021-05-04
Reorganize docs and change theme

prescod:
Before:

https://snowfakery.readthedocs.io/en/v1.10/

After:

https://snowfakery.readthedocs.io/en/feature-switch-theme/

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -44,9 +44,8 @@ So for example, if you had a Snowfakery file like this:
 - plugin: salesforce.org.DoGood
 ```
 
-named `do_goodders/do_gooder.recipe.yml`
-
-You could make a file named `do_gooders/plugins/salesforce/org/DoGood.py`
+named `do_goodders/do_gooder.recipe.yml`,
+you could make a file named `do_gooders/plugins/salesforce/org/DoGood.py`
 
 And that file would contain a class like this:
 
@@ -54,31 +53,45 @@ And that file would contain a class like this:
 from snowfakery.plugins import SnowfakeryPlugin
 
 class DoGood(SnowfakeryPlugin):
-    """Plugin which generates a summation helper"""
+    """Plugin which does good stuff"""
     ...
 ```
 
 ## Writing Plugins
 
-To write a new Plugin, make a class that inherits from `SnowfakeryPlugin` and implements either the `custom_functions()` method or a `Functions` nested class. The nested class is simple: each method represents a function to expose in the namespace. In this case the function name would be `DoublingPlugin.double`.
+To write a new Plugin, make a class that inherits from `SnowfakeryPlugin` and implements either the `custom_functions()` method or a `Functions` nested class. The nested class is simple: each method represents a function to expose in the namespace. In this case the
+namespace would be `DoublingPlugin` and the function name `double`:
+
+```yaml
+# examples/use_doubling_plugin.yml
+- plugin: mypackage.DoublingPlugin
+- object: sixer
+  fields:
+    six:
+      DoublingPlugin.double: 3
+```
 
 ```python
+# examples/plugins/mypackage/DoublingPlugin.py
+from snowfakery import SnowfakeryPlugin
+
+
 class DoublingPlugin(SnowfakeryPlugin):
     class Functions:
         def double(self, value):
             return value * 2
 ```
 
-Alternately, you can implement the custom_functions method to return an
-object with the attributes that implement your namespace:
+Alternately, you can implement the `custom_functions` method.
+
+In this case you return a Python object that implements your
+methods.
 
 ```python
 class Doubler:
   def double(self, value):
       return value * 2
-```
 
-```python
 class DoublingPlugin(SnowfakeryPlugin):
     def custom_functions(self, *args, **kwargs):
         return Doubler()
@@ -88,7 +101,7 @@ Make sure to accept `*args` and `**kwargs` to allow for future extensibility of 
 
 Despite the name, plugins can also include data values rather than functions in either form of plugin. Plugins essentially use Python's `getattr` to find attributes, properties, methods or functions in the namespace of the object you return from `custom_functions()`.
 
-Plugin functions can store persistent information in a Python dictionary called self.context.context_vars(). It will always be supplied to your plugin. For example, here is a simple plugin that counts:
+Plugin functions can store persistent information in a Python dictionary obtained by calling `self.context.context_vars()`. It will always be supplied to your plugin. For example, here is a simple plugin that counts:
 
 ```python
 class PluginThatCounts(SnowfakeryPlugin):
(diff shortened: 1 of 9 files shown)


docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+In the rare event that a plugin has a function which need its arguments to be passed to it unevaluated, for later (perhaps conditional) evaluation, you can use the `@snowfakery.lazy decorator`. Then you can evaluate the arguments with `self.context.evaluate()`.
+
+For example:
+
+```python
+class DoubleVisionPlugin(SnowfakeryPlugin):
+    class Functions:
+        @lazy
+        def do_it_twice(self, value):
+            "Evaluates its argument 0 times or twice"
```
davidmreed, line 124: How could it evaluate 0 times?

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+            rc = f"{self.context.evaluate(value)} : {self.context.evaluate(value)}"
+
+          return rc
```
davidmreed, line 127: indentation?

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+This would output an `OBJ` row with values:
+
+```python
+  {'id': 1, 'some_value': 'abc : abc', 'some_value_2': '1 : 2'})
+```
+
+Occasionally you might write a plugin which needs to evaluate its
+parameters lazily but doesn't care about the internals of the values
+because it just returns it to some parent context. In that case,
+use `context.evaluate_raw` instead of `context.evaluate`.
```
davidmreed, line 150:
```suggestion
use `context.evaluate_raw()` instead of `context.evaluate()`.
```

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+
+Plugins that require "memory" or "state" are possible using PluginResult
```
davidmreed, line 152:
```suggestion
Plugins that require "memory" or "state" are possible using `PluginResult`
```

docs/index.md
```diff
@@ -1020,45 +941,96 @@ The relevant section of the Jinja document is called  [Expressions](https://jinj
...
 
-Of course you can do any math you want in the formula:
+#### `today`
+
+The `today` variable returns a date
+representing the current date. This date
+will not chanage during the execution of
+a single recipe.
+
+#### `fake.`
```
davidmreed, line 974: Is this period spurious?

docs/salesforce.md
```diff
@@ -0,0 +1,162 @@
...
+    RecordType: HH_Account
+- object: Account
+  fields:
+    name: The Firm
+    RecordType: Organization
+```
+
+## Creating and Referencing Person Accounts
+
+There are several features planned for the Salesforce Plugin, but
```
davidmreed, line 78: This sentence makes it slightly unclear whether Person Account support is currently available or planned.

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+
+So for example, if you had a Snowfakery file like this:
+
+```yaml
+- plugin: salesforce.org.DoGood
+```
+
+named `do_goodders/do_gooder.recipe.yml`
+
+You could make a file named `do_gooders/plugins/salesforce/org/DoGood.py`
```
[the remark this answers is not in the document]
prescod: I don't think so.

https://stackoverflow.com/a/37140173/113477

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+- in a `plugins` directory below the current working directory and
+- in a subdirectory of the user's home directory called `.snowfakery/plugins`.
+
+So for example, if you had a Snowfakery file like this:
+
+```yaml
+- plugin: salesforce.org.DoGood
+```
+
+named `do_goodders/do_gooder.recipe.yml`
```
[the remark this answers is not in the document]
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-46aca7079be2ef59a99745c7589744a1171129649b89add013cb0c6e1b2bd8b7R47

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+
+```python
+class DoublingPlugin(SnowfakeryPlugin):
+    class Functions:
+        def double(self, value):
+            return value * 2
+```
+
+Alternately, you can implement the custom_functions method to return an
+object with the attributes that implement your namespace:
```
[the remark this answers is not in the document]
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-46aca7079be2ef59a99745c7589744a1171129649b89add013cb0c6e1b2bd8b7R85

docs/extending.md
```diff
@@ -0,0 +1,299 @@
...
+    class Functions:
+        @lazy
+        def do_it_twice(self, value):
+            "Evaluates its argument 0 times or twice"
+            rc = f"{self.context.evaluate(value)} : {self.context.evaluate(value)}"
+
+          return rc
+```
+
+Every second time this is called, it will evaluate its argument twice, and stick the two results into a string. For example, if it were called with a call to `random_number`, you would get two different random numbers rather than the same number twice. If it were called with the counter from above, you would get two different counter values in the string.
```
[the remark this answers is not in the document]
prescod: The example changed after I wrote that.

https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-46aca7079be2ef59a99745c7589744a1171129649b89add013cb0c6e1b2bd8b7R143

docs/extending.md
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-46aca7079be2ef59a99745c7589744a1171129649b89add013cb0c6e1b2bd8b7R137

docs/extending.md
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-46aca7079be2ef59a99745c7589744a1171129649b89add013cb0c6e1b2bd8b7R140

docs/salesforce.md
prescod: https://github.com/SFDO-Tooling/Snowfakery/pull/353/commits/63e125ab927e8427ae9f5956554f86e8b0e842f8#diff-444f7dedc7f85729133c0374d624e8dacaf5d5d4b3d068064765275f73306683R79

docs/index.md
prescod: No, just laziness. Clarified.

d26955c

davidmreed (approved):

PR #356, opened 2021-05-05
Dependency roll-up: May 5, 2021

prescod:


requirements_dev.txt
```diff
@@ -1,5 +1,5 @@
 -r requirements.txt
-black==20.8b1
+black==21.4b2
```
davisagli, line 2: black also needs to be updated here https://github.com/SFDO-Tooling/Snowfakery/blob/main/.pre-commit-config.yaml#L5 and run on existing files

davisagli (approved):

PR #365, opened 2021-05-07
Snowbench: Usability improvements suggested in code review

prescod:
1. Use :n format specifier to get locale-friendly behaviours
2. Supply richer default arguments.
3. Use click.progressbar
4. Fix a SQL Alchemy warning about a deprecated method
5. Test defaults in CI.


snowfakery/tools/snowbench.py
```diff
@@ -40,9 +47,21 @@ def snowbench(
     execution process is spawned off into a sub-process instead
     of being executed inline.
     """
-    with TemporaryDirectory() as tempdir:
+    with TemporaryDirectory() as tempdir, click.progressbar(
+        label="Benchmarking", length=num_records
```
jstvz, line 51:
```suggestion
        label="Benchmarking", length=num_records, show_eta=False
```

The ETA is off by several orders of magnitude, so I suggest we hide it.

jstvz (dismissed):

jstvz (approved):

Issue #282, opened 2021-03-10
Allow for multiple combinations of selections (multi-select picklist)


Julian88Tex:
### Description
Add in the ability to allow for multiple selections in order to work with multi-select picklists or similar database functionality. Currently, `random_choice` only allows for one individual selection at a time. Multi-select picklists can have multiple selections.

@prescod gave the example of DNA and proposed a new `random_choices` function:

```
DNA: 
  random_choices:
   min: 0
   max: 5_000_000
   with_replacement: True
   choices:
      - A
      - C
      - G
      - T
```

### Above Potential Solution Description
* `min` would be the minimum number of potential values returned which could be 0 (null). If no min is specified then `1` is assumed.
* `max` would be the maximum number of potential values returned which could be any number. If no max is specified then the number of choices is assumed to be the max. The above example would default to `4`.

In the above-proposed solution acceptable possible results would be:
- "ACT"
- "CGT"
- "AT"
- "AAAACCCGGTAACGT
- ""
- "G"

Additional info for Salesforce Internal Users is available here: https://salesforce-internal.slack.com/archives/C01KBFT16Q2/p1614027103002800

Julian88Tex:
@prescod Would you mind looking over this issue? Not sure if I captured everything you were thinking correctly. Also, couldn't remember what `with_replacement` meant.

prescod:
@Julian88Tex What is the best SFDO product to install to experiment with multi-select picklists? Or is there a standard object field I could use?

Julian88Tex:
@prescod I know of an SFDO-ish product from Salesforce Impact Labs project. I have [this branch](https://github.com/Salesforce-org-Impact-Labs/01HousingandHomelessness/blob/feature/snowfakery-dataset/datasets/snowfakery.yml) which I was working on a Snowfakery yml which includes [this multiselect picklist fied](https://github.com/Salesforce-org-Impact-Labs/01HousingandHomelessness/blob/feature/snowfakery-dataset/force-app/main/default/objects/Service__c/fields/All_Types__c.field-meta.xml).

umeditor:
I ran into this when trying to populate EDA's hed__Race__c field, a multi-select picklist. As people can have more than one race, it would be great if Snowfakery could select more than one option. Right now, I'm just doing:

```
hed__Race__c:
  random_choice:
    choices:
        - American Indian/Alaskan Native
        - Asian
        - Black/African American
        - Native Hawaiian/Other Pacific Islander
[...]

PR #371, opened 2021-05-10
Experimental support for multi-select picklists

prescod:
Experimental and interim support for multi-select picklists.


prescod:
Related to #282

PR #372, opened 2021-05-11
Update requirements

prescod:


requirements_dev.txt
```diff
@@ -1,13 +1,13 @@
...
-flake8==3.9.1
-myst-parser==0.13.7
+flake8==3.9.2
+myst-parser==0.14.0
 pre-commit==2.12.1
-pytest==6.2.3
+pytest==6.2.4
 pytest-cov==2.11.1
-Sphinx==3.5.4
+Sphinx==3.5.4      # pin this until Myst is happy with Sphinx 4
```
davisagli, line 10: @Br4nd0R could you help Paul get snowfakery switched to pip-tools so we have a better way to express this kind of thing?

davisagli (approved):

PR #388, opened 2021-05-25
Pass kwargs to Faker

prescod:
Fix regression where function arguments were not passed to faker properly.

### DIFF
--- a/snowfakery/fakedata/fake_data_generator.py
+++ b/snowfakery/fakedata/fake_data_generator.py
@@ -46,7 +46,7 @@ class FakeData:
         else:
             try:
                 # TODO: look for FooBar as foo_bar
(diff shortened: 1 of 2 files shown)


davisagli (commented):
Does faker ever use positional args?

prescod:
Faker does not seem to require positional args and all example code uses keyword-arg syntax. I'm comfortable with enforcing that in Snowfakery.

davisagli:
@prescod My question is more whether there are existing recipes that call faker with positional args, which we would break.

prescod:
It's possible that someone discovered that syntax by themselves, but I haven't seen it. The would have had their code broken 2 weeks ago but they might not have upgraded. I can put a note in the release notes.

davisagli (approved):

PR #389, opened 2021-05-25
Dependabot rollup

prescod:


requirements.txt
```diff
@@ -1,9 +1,9 @@
-SQLAlchemy==1.4.14
-Faker==8.1.2
-jinja2==2.11.3
+SQLAlchemy==1.4.15
+Faker==8.2.1
+jinja2==3.0.1
 PyYAML==5.4.1
-click==7.1.2
+click==8.0.1
```
davisagli, line 5: cumulusci's requirements/prod.in currently specifies `click<8` and `jinja2<3` which will conflict with this. That probably means that @Br4nd0R tried updating them but saw errors (either compatibility with other requirements, or test failures). Could you either take a closer look or keep these back at the versions that cumulusci is using?

I'm trying to think how we could handle this more automatically. We could add a Snowfakery build which fetches the main branch of cumulusci and at least confirms pip-compile can resolve dependencies that are compatible with both snowfakery and cumulusci. Or we could use cumulusci's requirements as a constraint file when compiling snowfakery's requirements using pip-compile.

davisagli (changes requested):

requirements.txt
prescod: For now I've rolled them back and I can work with @Br4nd0R on moving to pip-tools to implement the other ideas.

davisagli (approved):

PR #403, opened 2021-06-10
Allow names to use a Salesforce naming convention

prescod:
Allow names to use a Salesforce naming convention: 

phone_number -> PhoneNumber

Alternate spellings are all valid:

phonenumber
PhoneNumber
phone_number (original)
PHONENUMBER

In the unlikely case of there being names ab_cd and abc_d, the caller can use underscores to disambiguate.

### DIFF
--- a/snowfakery/fakedata/fake_data_generator.py
+++ b/snowfakery/fakedata/fake_data_generator.py
@@ -25,6 +25,8 @@ class FakeNames(T.NamedTuple):
         return self.f.email()
 
 
+# we will use this to exclude Faker's internal book-keeping methods
+# from our faker interface
 faker_class_attrs = set(dir(Faker))
 
 
--- a/snowfakery/fakedata/fake_data_generator.py
+++ b/snowfakery/fakedata/fake_data_generator.py
@@ -26,32 +26,41 @@ class FakeNames(T.NamedTuple):
 
 
 class FakeData:
-    """Wrapper Faker which adds Salesforce names and case insensitivity."""
+    """Wrapper for Faker which adds Salesforce names and case insensitivity."""
 
     def __init__(self, faker: Faker):
-        self.faker = faker
-        self.fake_names = FakeNames(faker)
+        faker = faker
+        fake_names = FakeNames(faker)
 
-        # canonical form of names is lower-case
-        self._lowers = {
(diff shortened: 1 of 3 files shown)


snowfakery/fakedata/fake_data_generator.py
```diff
@@ -26,32 +26,41 @@ def RealisticMaybeRealEmail(self):
 
 
 class FakeData:
-    """Wrapper Faker which adds Salesforce names and case insensitivity."""
+    """Wrapper for Faker which adds Salesforce names and case insensitivity."""
 
     def __init__(self, faker: Faker):
-        self.faker = faker
-        self.fake_names = FakeNames(faker)
+        faker = faker
```
davisagli, line 32: ?

snowfakery/fakedata/fake_data_generator.py
```diff
@@ -26,32 +26,41 @@ def RealisticMaybeRealEmail(self):
...
+        fake_names = FakeNames(faker)
 
-        # canonical form of names is lower-case
-        self._lowers = {
-            v.lower(): v for v in dir(self.fake_names) if not v.startswith("_")
+        def obj_to_func_list(obj: object, canonicalizer: T.Callable):
+            return {
+                canonicalizer(name): getattr(obj, name)
+                for name in dir(obj)
+                if not name.startswith("_") and name != "seed"
```
davisagli, line 39: What is "seed"?

snowfakery/fakedata/fake_data_generator.py
prescod: That was "self.faker = faker" and then I did a search and replace to get rid of "self.faker".

snowfakery/fakedata/fake_data_generator.py
prescod: Is it clearer in 975d5f2?

davisagli: Not really. I knew we were ignoring "seed" but not why. Now it doesn't mention "seed" at all and I'm not sure if that's an oversight in the refactoring or what.

prescod: Okay trying to clarify again: fc91d7e

davisagli (approved):

PR #404, opened 2021-06-10
Make CumulusCI Optional

prescod:
Make CumulusCI an optional requirement of Snowfakery. If you don't have it installed, some tests will be skipped.


davisagli (changes requested):
Shouldn't this also include a change to .github/workflows/ci.yml to make sure that cumulusci is installed there?

davisagli:
What's up with the test failure?

prescod:
#405 fixes the test failures, @davisagli

davisagli (approved):

PR #438, opened 2021-07-09
Feature/unique

prescod:
The `unique_id` variable returns a unique number that can be used to distinguish any record from others. For example, we can incorporate a unique ID into an email address or an employee ID.

The rest of its features are documented below. It is also implemented by an undocumented Plugin which might be independently useful at some point (e.g. multiple uniqueness generators in the same program).

Future: allow alphanumeric IDs based on python-baseconv.


examples/plugins/RecipeStateDemo.py
```diff
@@ -0,0 +1,14 @@
+from snowfakery import SnowfakeryPlugin
+
+
+class RecipeStateDemo(SnowfakeryPlugin):
+    class Functions:
+        def show_recipe_state(self, _):
+            print(dir(self.context))
+            breakpoint()
```
jofsky, line 8: @prescod this intentional?

docs/index.md
```diff
@@ -866,13 +904,77 @@ executed again.
...
+# examples/unique_id/test_unique_id.yml
+- object: Contact
+  count: 3
+  fields:
+    FirstName:
+      fake: FirstName
+    LastName:
+      fake: LastName
+    EmployeeNum: ${{unique_id}}
+    Email: ${{unique_id}}_${{fake.email}}
```
jofsky, line 924: @prescod is `fake` shared for FirstName and LastName Object instantiations? How does it know what email to get?

tests/plugins/test_unique_id.py
```diff
@@ -0,0 +1,330 @@
...
+            StringIO(yaml),
+            target_number=("Example", 4),
+            plugin_options={"big_ids": "False"},
+        )
+        assert generated_rows.row_values(0, "unique")
+        assert generated_rows.row_values(1, "unique")
+        assert generated_rows.row_values(2, "unique")
+        assert generated_rows.row_values(3, "unique")
+
+    # def test_continuations(self, generate_data_with_continuation, generated_rows):
```
jofsky, line 40: Should we remove this?

jofsky (approved):

docs/index.md
prescod: The email is faked too. 

But yes it does have some internal heuristics which try to make the email match the firstname and lastname based on shared state attached to the Snowfakery context, however.

jofsky (approved):

PR #441, opened 2021-07-14
More forgiving parsing of functions

prescod:
Snowfakery represents functions with keyword arguments as a dict with a single key/value pair like this:

```
choice:
            when: ${{HomePhone!=NULL}}
            pick: Home
```

Which is the YAML equivalent of:

```
{choice: {"when": "${{HomePhone!=NULL}}", "pick", "Home"}}
```

Because Snowfakery often left-align keys, users often try to type:

```
choice:
when: ${{HomePhone!=NULL}}
pick: Home
```
This patch allows that syntax.


jkasturi-sf (approved):
Looks good

PR #442, opened 2021-07-15
More tests for Salesforce Stuff

prescod:
Unit test for some of the Salesforce query features.

### DIFF
--- a/tests/test_salesforce_gen.py
+++ b/tests/test_salesforce_gen.py
@@ -1,6 +1,9 @@
 from base64 import b64decode
 
+import pytest
 from snowfakery import generate_data
+from snowfakery.standard_plugins.Salesforce import SalesforceConnection
+from snowfakery import data_gen_exceptions as exc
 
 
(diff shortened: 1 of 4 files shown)


tests/test_salesforce_gen.py
```diff
@@ -11,3 +14,12 @@ def test_content_version(self, generated_rows):
         rawdata = b64decode(b64data)
         assert rawdata.startswith(b"%PDF-1.3")
         assert b"Helvetica" in rawdata
+
+
+class TestSalesforceConnection:
+    def test_bad_kwargs(self):
+        sfc = SalesforceConnection(None)
+        with pytest.raises(exc.DataGenError, match="Unknown argument"):
```
Br4nd0R, line 22: Is the unknown argument `xyzzy` or is it the `blah` field? It would be nice if the error message included the name of the unknown argument.

Br4nd0R (changes requested):

PR #469, opened 2021-08-07
Better snowfakery version message

prescod:
Changes:

Clearer Snowfakery version error.

### DIFF
--- a/snowfakery/cli.py
+++ b/snowfakery/cli.py
@@ -3,7 +3,7 @@ import sys
 from pathlib import Path
 
 from snowfakery.data_gen_exceptions import DataGenError
-
+from snowfakery.utils.versions import check_latest_version
 
 import click
 from snowfakery import version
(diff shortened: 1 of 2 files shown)


snowfakery/cli.py
```diff
@@ -51,6 +51,33 @@ def int_string_tuple(ctx, param, value=None):
...
+        yaml = """
+        - object: Status
+          fields:
+            quote: Shiny and new
+        """
+        out = StringIO()
+        generate_data(StringIO(yaml), output_files=[out], output_format="txt")
+        return "Properly installed" if out.getvalue() else "Unknown installation error"
+
+    def __mod__(self, vals) -> str:
```
davisagli, line 67: This needs a comment explaining how it works, and a test that will break if `click` switches to a different way of formatting the string.

davisagli (changes requested):
Needs tests

PR #477, opened 2021-08-24
Support all relevant mapping features in load.yml

prescod:
Snowfakery-in-CCI now allows users to set the batch_size and anchor_date for dataloads.

### DIFF
--- a/snowfakery/cci_mapping_files/declaration_parser.py
+++ b/snowfakery/cci_mapping_files/declaration_parser.py
@@ -28,7 +28,7 @@ class AtomicDecl(T.NamedTuple):
 
     sf_object: str
     key: str
-    value: T.Union[T.List, str]
+    value: T.Union[T.List, str, date, int]
     priority: int
     merge_rule: T.Callable  # what to do if two declarations for same val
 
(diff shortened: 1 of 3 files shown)


snowfakery/cci_mapping_files/declaration_parser.py
```diff
@@ -28,7 +28,7 @@ class AtomicDecl(T.NamedTuple):
 
     sf_object: str
     key: str
-    value: T.Union[T.List, str]
+    value: T.Union[T.List, str, date, int]
```
davisagli, line 31: If this is used for parsing, I don't think it will do what you want. In a Union, pydantic uses the first type that the input can successfully be cast to. I think it'd be a good idea to make the test_cli__explicit_file test assert that the values loaded from the file are in the expected format.

davisagli (changes requested):

snowfakery/cci_mapping_files/declaration_parser.py
davisagli: (Oh, but it probably works because it's parsed by yaml first?)

prescod: Sorry for the long delay. This is a NamedTuple not a Pydantic Model, so your concern does not arise.

davisagli (approved):

davisagli (commented):
Docs needed?

PR #485, opened 2021-09-08
Add 'now' variable to Snowfakery

prescod:
The `now` variable returns a datetime representing the current moment. It stores microsecond precision.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -983,7 +983,42 @@ Contact(id=3, FirstName=Gene, LastName=Wall, DepartmentCode=42Q3XX3N)
 
 #### `today`
 
-The `today` variable returns a date representing the current date. This date does not change chanage during the execution of a single recipe.
+The `today` variable returns a date
+representing the current date. This date
+will not chanage during the execution of
+a single recipe. See the `date` function
+to learn more about this type of object.
+
+#### `now`
+
+The `now` variable returns a [datetime](#datetime)
+representing the current moment. It stores
+microsecond precision.
+
+Using this recipe, you can see three different
+ways of outputting the timestamp:
+
+```yaml
+# tests/test_now.yml
+- object: Times
+  fields:
+    current_datetime: ${{now}}
+    current_datetime_as_number: ${{now.timestamp()}}
+    current_datetime_without_microseconds: ${{int(now.timestamp())}}
+```
+
+This would generate field values similar to these:
+
+```yaml
+current_datetime="2021-09-08 15:12:47.444352"
+current_datetime_as_number=1631139167.444973
+current_datetime_without_microseconds=1631139167
+```
+
+Experimentally, this variable seems to return a unique value
+every time it is called, but it might depend on
+your operating system and hardware setup
+(e.g. a very fast CPU with a very slow system clock).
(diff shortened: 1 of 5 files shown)


tests/test_template_funcs.py
```diff
@@ -241,6 +242,20 @@ def test_date_from_datetime(self, write_row):
...
 
+    def test_now_variable(self, generated_rows):
+        yaml = """
+        - object : A
+          fields:
+            a: ${{now}}
+            b: ${{now}}
+        """
+        generate(StringIO(yaml), {}, None)
+        assert datetime.fromisoformat(generated_rows.table_values("A", 0, "a"))
```
boakley, line 253: This test seems a little imprecise -- it's verifying that `now` generates _a_ date, but it doesn't verify that it sets the correct date. Would it be worth taking the time to mock out datetime.now?

docs/index.md
```diff
@@ -933,6 +933,19 @@ representing the current date. This date
 will not chanage during the execution of
 a single recipe.
 
+#### `now`
+
+The `now` variable returns a datetime
+representing the current moment. Experimentally,
```
boakley, line 939: Would it make sense to specifically mention that this returns the value of `datetime.now()`? I don't don't if the target audience is savvy enough to know what that means.

boakley (dismissed):
I left a couple of small comments, but I'm going to go ahead and approve because I don't think my questions are showstoppers.

tests/test_template_funcs.py
prescod: @boakley Fixed in 9329a0d8cc83c66eb0a5d97414124e7c33b36384

docs/index.md
prescod: I think not, in general. I'll see what Peter says.

prescod:
I addressed David's suggestion that the date have a timezone. Since this PR was created, Snowfakery adopted a UTC-default policy.

docs/index.md
```diff
@@ -983,7 +983,40 @@ Contact(id=3, FirstName=Gene, LastName=Wall, DepartmentCode=42Q3XX3N)
 
 #### `today`
 
-The `today` variable returns a date representing the current date. This date does not change chanage during the execution of a single recipe.
+The `today` variable returns a date
+representing the current date. This date
+will not chanage during the execution of
```
pconrad-sfdo, line 988:
```suggestion
doesn't change during the execution of
```

docs/index.md
```diff
@@ -983,7 +983,40 @@ Contact(id=3, FirstName=Gene, LastName=Wall, DepartmentCode=42Q3XX3N)
...
+```yaml
+# tests/test_now.yml
+- object: Times
+  fields:
+    current_datetime: ${{now}}
+    current_datetime_as_number: ${{now.timestamp()}}
+    current_datetime_without_microseconds: ${{int(now.timestamp())}}
+```
+
+This would generate field values similar to these:
```
pconrad-sfdo, line 1010:
```suggestion
The above example generates field values similar to these:
```

boakley (approved):
Looks good to me.

PR #497, opened 2021-09-27
Parent child just once bugfix

prescod:
Previously, the `random_reference` feature could not refer to objects created with just_once.  This has been corrected.


jofsky (approved):

davisagli:
Is the failed "with cumulusci" build a problem?

prescod:
@davisagli 95% chance it is caused by version skew between our fixtures. I think it's safe to proceed because tests pass on my computer when I use these branches with CCI main.

PR #500, opened 2021-10-01
Allow bools and dates as options for random_choice's short syntax

prescod:
random_choice has a short-form key:probability syntax. Now that short-form syntax can be used for booleans and dates, not just strings.

### DIFF
--- a/snowfakery/parse_recipe_yaml.py
+++ b/snowfakery/parse_recipe_yaml.py
@@ -147,7 +147,6 @@ def removeline_numbers(dct: Dict) -> Dict:
 
 
 def _coerce_to_string(val, context):
-    print(val, type(val))
     if isinstance(val, str):
         return val
     elif isinstance(val, (int, bool, date)):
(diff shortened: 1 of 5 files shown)


snowfakery/parse_recipe_yaml.py
```diff
@@ -146,14 +146,34 @@ def removeline_numbers(dct: Dict) -> Dict:
     return {i: dct[i] for i in dct if i != "__line__"}
 
 
+def _coerce_to_string(val, context):
+    print(val, type(val))
```
boakley, line 150: Was this print statement intentionally left in?

prescod: Oops, no, fixed in a0fb7ef.

prescod:
@boakley i fixed the issue you noticed, but it triggered a bug in `coverage`. The workaround is to improve coverage so I added an unrelated test to improve coverage.

boakley (commented):
Looks good to me.

PR #541, opened 2021-11-19
Snowfakery docs

prescod:
@pconrad-sfdo  This PR is probably too big for a meaningful review. It would make more sense to just rubber stamp it and then plan for a top-to-bottom Snowfakery docs review when you have time. I mean it's up to you, but reviewing 3/4 of the docs in "diff" format seems harder than reviewing 100% of it in Markdown.


pconrad-sfdo:
@prescod I agree. If it's been reviewed multiple times, that implies a certain level of quality. I'm happy to be part of the "rubber stamp" approach, and then we can take a step back and see what refinements we'd like.

pconrad-sfdo (approved):

Issue #479, opened 2021-08-27
Fake.PostalCode can generate 4 digit codes without leading zeroes (which are invalid)


dcinzona:
4 digit postal codes are invalid.  The generator should support leading zeroes.

https://github.com/joke2k/faker/blob/a565d46e928dba18938164de2125de1aabc3f825/faker/providers/address/en_US/__init__.py#L281

But `Faker` accounts for len==4
https://github.com/joke2k/faker/blob/a565d46e928dba18938164de2125de1aabc3f825/faker/providers/address/en_US/__init__.py#L386-L394

Unfortunately, I cannot tell if state_abbr is passed into the faker generator

allisonletts:
This issue should be closed, I think 👍

PR #551, opened 2021-11-29
Allow zero-prefixed strings of digits

prescod:
Snowfakery infers integers from digit strings in Jinja templates. The old pattern would strip leading zeros from digit strings starting with 0. (01234 => 1234). This is especially problematic for US postal codes as discovered in #479.

This patch leaves 0-prefixed strings as strings.

### DIFF
--- a/snowfakery/utils/template_utils.py
+++ b/snowfakery/utils/template_utils.py
@@ -66,7 +66,7 @@ number_chars = set(string.digits + ".")
 
 def look_for_number(arg):
     looks_like_float = False
-    if len(arg) == 0 or arg[0] == "0":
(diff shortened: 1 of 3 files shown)


davisagli (changes requested):
How would I enter the float `0.1`?

codexmarcel (dismissed):

davisagli (dismissed):
Any note needed in the docs?

prescod:
@davisagli I think it's just a bugfix, because no "real" integer should be impacted. So I'll mention in the release notes.

Also, please Approve again because I did some benchmarking and changed the code slightly.

davisagli (approved):

PR #555, opened 2021-12-01
Do not search locales and do not document "fake: zipfile" and similar irrelevant fakers

prescod:
There are two things in this branch:

1. Config for mkdocs search index build to prevent extremely slow searching:
    https://snowfakery.readthedocs.io/en/latest/search.html?q=postcode
versus:
    https://snowfakery.readthedocs.io/en/feature-dont-search-locales/search.html?q=postcode

2. While I was researching I noticed that the index is bloated by Fake data for things like "fake tarfile", "take zipfile" etc. Now I have a feature for ignoring those.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,6 +19,10 @@ markdown_extensions:
       toc_depth: 4
 plugins:
   - search
+  - exclude-search:
+      exclude:
+        - docs/fakedata/*.md
+        - fakedata/*.md
   - snowfakery_fakes:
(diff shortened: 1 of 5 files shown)


mkdocs.yml
```diff
@@ -19,6 +19,10 @@ markdown_extensions:
       toc_depth: 4
 plugins:
   - search
+  - exclude-search:
```
Br4nd0R, line 22: TIL

Br4nd0R (approved):

PR #556, opened 2021-12-01
Publish a JSON Schema

prescod:
- Add a JSON Schema for Snowfakery
- Add an example file for testing
- Add new snowcheck tool
- Delete a file that was accidentally checked in at some point
-


jstvz (approved):
LGTM

PR #563, opened 2021-12-14
Fix load order logic in the case of self-recursion

prescod:
Previously, if an object directly referenced itself (e.g. Users referencing each other in a Manager/Employee relationship), Snowfakery would treat it as being part of an unresolvable circular dependency and fairly randomly select the order. Now it will ignore self-references and use other cues to determine the load order.


davisagli (approved):
Makes sense to me

PR #565, opened 2021-12-21
Improve error message of an unusual corner case

prescod:
I finally hit this code path and cleaned it up.


davisagli:
@prescod could you explain what the scenario is so there's a bit of context for looking at the code?

pconrad-sfdo (approved):

pconrad-sfdo:
LGTM

prescod:
@davisagli :

In a statement list, I might classify something as a variable or object because it has a key named 'var' or 'object'. But there is no actual value that serves as the variable or object's "name".

https://github.com/SFDO-Tooling/Snowfakery/pull/565/files#diff-237d6caeea59e3cf8f303958b4b29e752861048496a047a6af38a12915a02e32R54

pconrad-sfdo (approved):

davisagli (approved):

PR #566, opened 2021-12-22
Remove uses of sets() which make iteration order unpredictable.

prescod:
A user reported that in some cases her recipe loaded okay and in other cases it failed. The underlying cause of the failed loads is that there are two tables that had an implicit, not explicit, dependency. A picklist populated by CampaignMemberStatus had to be loaded before CampaignMembers could be.

We already have a way to force a load order, but my concern was that the intermittent behaviour made tracking down the source of the problem harder.

Also: the user's mental model was that the load order would follow the recipe.

AC:

This change ensures that the load order does actually follow the recipe except if a dependency can be inferred. Dependency order takes priority over recipe declaration order.


davisagli:
What work item is this related to?

prescod:
It's related to an issue reported by users in Slack.

docs/salesforce.md
```diff
@@ -51,6 +50,14 @@ $ cci flow run test_everything
 ...
 ```
 
+The order in which sobjects is loaded is inferred by Snowfakery and CumulusCI
```
davisagli, line 53:
```suggestion
The order in which sObjects are loaded is inferred by Snowfakery and CumulusCI
```

snowfakery/utils/collections.py
```diff
@@ -67,3 +67,11 @@ def copy(self):
 
     def __repr__(self):
         return str(dict(self.items()))
+
+
+class SortedSet(OrderedDict):
```
davisagli, line 72: How about calling it OrderedSet? SortedSet makes it sound like it will do the equivalent of sort() as its contents are mutated.

Could it simply extend `dict` now that we dropped support for Python versions where dict is not ordered?

docs/salesforce.md
pconrad-sfdo:
```suggestion
Snowfakery and CumulusCI infer the order in which sObjects are loaded
```

prescod:
Text is now:

Snowfakery and CumulusCI infer the order in which sObjects are loaded
based primarily on which tables have fields that depend on which other
tables.

davisagli (approved):

PR #590, opened 2022-01-26
Allow parent applications to create output streams and fix datetime

prescod:
When Snowfakery is embedded in CumulusCI it needs to format data slightly differently because Salesforce requires date-times to have a timezone. This puts the parent application (like CumulusCI) in charge of supplying OutputStreams, so it can capture information however it wants. Perhaps one day CumulusCI will be able to start sending data to Salesforce before Snowfakery has finished executing even a single iteration of the recipe.

Comments explain the changes.

Note for HISTORY.rst: recipes that depended on fakers always being serialized to strings will need to do it explicitly now.

### DIFF
--- a/snowfakery/api.py
+++ b/snowfakery/api.py
@@ -114,6 +114,12 @@ class SnowfakeryApplication:
 
         return last_used_id >= target_id
 
+    def configure_output_stream(
+        self,
+        **kwargs,
+    ):
+        return configure_output_stream(parent_application=self, **kwargs)
+
 
 def stopping_criteria_from_target_number(target_number):
     "Deconstruct a tuple of 'str number' or 'number str' and make a StoppingCriteria"
@@ -163,12 +169,15 @@ def generate_data(
             return exit_stack.enter_context(open_file_like(file, mode))
 
         parent_application = parent_application or SnowfakeryApplication(
-            stopping_criteria
+            stopping_criteria,
         )
 
         output_stream = exit_stack.enter_context(
-            configure_output_stream(
-                dburls, output_format, output_files, output_folder, parent_application
+            parent_application.configure_output_stream(
+                dburls=dburls,
+                output_format=output_format,
+                output_files=output_files,
+                output_folder=output_folder,
             )
         )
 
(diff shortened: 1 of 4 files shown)


snowfakery/api.py
```diff
@@ -114,6 +114,12 @@ def check_if_finished(self, id_manager):
 
         return last_used_id >= target_id
 
+    def configure_output_stream(
```
prescod, line 117: Puts the parent application in charge of the output stream creation. By default it creates the same output streams it always did.

snowfakery/data_generator_runtime.py
```diff
@@ -569,7 +569,11 @@ def executable_blocks(self):
         return {**self.field_funcs(), "fake": self.fake}
 
     def fake(self, name):
-        return str(self.runtime_context.faker_template_library._get_fake_data(name))
+        val = self.runtime_context.faker_template_library._get_fake_data(name)
```
prescod, line 572: Previously the faker function serialized everything to strings. Now it does not. The user can still do so if they want to, by using jinja filters.

snowfakery/output_streams.py
```diff
@@ -125,7 +124,7 @@ def close(self, **kwargs) -> Optional[Sequence[str]]:
 
         Return a list of messages to print out.
         """
-        return super().close()
+        pass
```
prescod, line 127: Discovered that OutputStreams that did not override close() would crash so I fixed that bug.

snowfakery/output_streams.py
```diff
@@ -278,9 +277,7 @@ class SqlDbOutputStream(OutputStream):
 
     should_close_session = False
 
-    def __init__(self, engine: Engine, mappings: None = None, **kwargs):
```
prescod, line 281: Old deprecated cruft removed.

snowfakery/output_streams.py
```diff
@@ -289,9 +286,7 @@ def __init__(self, engine: Engine, mappings: None = None, **kwargs):
         self.base = automap_base(bind=engine, metadata=self.metadata)
 
     @classmethod
-    def from_url(cls, db_url: str, mappings: None = None):
```
prescod, line 292: Old deprecated cruft removed.

PR #593, opened 2022-01-27
Clean up a warning

prescod:
Capture a warning in a test case instead of printing it to the console.

### DIFF
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -24,7 +24,7 @@ class TestEmbedding:
             assert dbpath.exists()
 
     def test_arguments(self):
-        with TemporaryDirectory() as t:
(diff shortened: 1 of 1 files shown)


Br4nd0R (approved):

tests/test_embedding.py
```diff
@@ -24,7 +24,7 @@ def test_embedding_dburl(self):
             assert dbpath.exists()
 
     def test_arguments(self):
-        with TemporaryDirectory() as t:
+        with TemporaryDirectory() as t, mock.patch("warnings.warn") as w:
```
davisagli, line 27: You could have used https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings instead of mock, which looks like it would make it straightforward to assert that the expected warning was shown

PR #594, opened 2022-01-27
Check for case where CSV Row has too many columns

prescod:
Better error checking for the case where a CSV Row has too many columns.

### DIFF
--- a/snowfakery/standard_plugins/datasets.py
+++ b/snowfakery/standard_plugins/datasets.py
@@ -9,7 +9,7 @@ from sqlalchemy.sql.expression import func, select
 from sqlalchemy.sql.elements import quoted_name
 
 from yaml.representer import Representer
-from snowfakery.data_gen_exceptions import DataGenNameError
+from snowfakery.data_gen_exceptions import DataGenError, DataGenNameError
 
 from snowfakery.plugins import (
     SnowfakeryPlugin,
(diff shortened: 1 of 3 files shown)


jofsky (approved):
Add docstrings!

snowfakery/standard_plugins/datasets.py
```diff
@@ -113,12 +113,22 @@ def __init__(self, datasource: Path, repeat: bool):
...
+        self.results = (plugin_result(row) for row in d)
 
     def close(self):
         self.results = None
         self.file.close()
 
+    def plugin_result(self, row):
+        if None in row:
+            raise DataGenError(
+                f"Your CSV row has more columns than the CSV header:  {row[None]}, {self.datasource}"
```
davisagli, line 127: Would be helpful to indicate which row had the problem

prescod:
@*davisagli* : The row[None] stuff outputs the "extra values" so you can
search for them.

davisagli:
@prescod Yeah I understand what it does, and it doesn't sound as helpful as a line number if the extra values happen to also appear in legitimate column positions.

prescod:
I want to add as few instructions to the golden path as possible. It's a
pretty rare situation. I was already reluctant to add
additional instructions to the golden path to deal with it. You would have
really hated my first solution which was zero overhead but gave even less
context. I think this is a good middle ground between making every user pay
for the error handling and giving enough information for the error case to
be discoverable. If you load your CSV into Sheets or Excel, the extra row
should be visually obvious.

davisagli:
It's hard to imagine that incrementing a counter would affect performance noticeably, but the concern with performance here seems fair.

PR #596, opened 2022-02-01
Move hidden field refactoring under control of Interpreter

prescod:
Move hidden field refactoring under control of Interpreter, not OutputStream

This will facilitate a feature of the Updates branch such that fields generated by Updates do not output "id" column.

"Regular" Snowfakery does always output "id" automatically. But updates should only output "id" if explicitly asked for. So I need to turn it off at the interpreter level, depending on the command line or the input from CumulusCI.

So I've moved the "filtering" code up into the interpreter and I'll add an option for filtering ids in the Updates branch.


davidjray (approved):
LGTM

PR #601, opened 2022-02-03
Datetimes have a UTC timezone by default

prescod:
Overrides Faker's DateTime-generating methods to return DateTimes with timezone set to UTC by default. Other timezones can be specified as a `relativedelta`.  Timezone handling can be turned off by specifying `False`.


tests/conftest.py
```diff
@@ -38,10 +37,17 @@ def table_values(tablename, index, field=None):
                 table = row[1][0]
                 mockobj._index.setdefault(table, []).append(row[1][1])
 
-        if field:  # return just one field
-            return mockobj._index[tablename][index][field]
-        else:  # return a full row
-            return mockobj._index[tablename][index]
+        if index is None:  # return all rows
```
prescod, line 40: These are some improvements to the testing infrastructure to make certain kinds of assertions easier to write. It doesn't require a deep review because it isn't user-facing code and it gets well-tested by virtue of being a test suite.

snowfakery/output_streams.py
```diff
@@ -35,6 +35,11 @@ def noop(x):
     return x
 
 
+def format_datetime(dt: datetime.datetime):
+    """Format into the Salesforce-preferred syntax."""
+    return str(dt).replace(" ", "T")
```
jstvz, line 40: What about `isoformat()` instead?
```suggestion
    return dt.isoformat(timespec="seconds")
```

tests/conftest.py
jstvz: 🔬 🕵🏾‍♂️

tests/test_faker.py
```diff
@@ -123,6 +123,65 @@ def test_months_past(self, write_row_mock):
         assert (date.today() - the_date).days > 80
         assert (date.today() - the_date).days < 130
 
+    def test_date_times(self, generated_rows):
+        with open("tests/test_datetimes.yml") as yaml:
+            generate(yaml)
+
+        for dt in generated_rows.table_values("Contact", field="EmailBouncedDate"):
+            assert "+0" in dt or dt.endswith("Z"), dt
+            assert dateparser.parse(dt).tzinfo
```
jstvz, line 132: What about using `dateparser.isoparse()` instead? `parse` is pretty liberal about what it accepts.

jstvz (approved):

snowfakery/data_generator_runtime.py
```diff
@@ -585,7 +585,11 @@ def executable_blocks(self):
         return {**self.field_funcs(), "fake": self.fake}
 
     def fake(self, name):
-        return str(self.runtime_context.faker_template_library._get_fake_data(name))
+        val = self.runtime_context.faker_template_library._get_fake_data(name)
+        if isinstance(val, Scalar.__args__):
```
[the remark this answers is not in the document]
prescod: I didn't do it exactly that way but: b622e7c

tests/test_faker.py
prescod: 09f197b

snowfakery/data_generator_runtime.py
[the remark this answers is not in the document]
jstvz: Did you mean to replace this with `ScalarTypes`?

jstvz (approved):
You've got `Scalar.__args__` [here](https://github.com/SFDO-Tooling/Snowfakery/pull/601/files#diff-7d6c33898abd6bd4043ac6eb4424877ad63c13af80eff3e6fd40e0e22be849bcR589) (if you meant to change it), but LGTM otherwise.

Issue #630, opened 2022-03-21
Decimals generated by Faker cause errors


prescod:
"No encoder found for <class 'decimal.Decimal'>" error is generated with fake.latitude

PR #631, opened 2022-03-22
Allow decimal datatype

prescod:
Decimal fakers were causing a problem because there was no serialization rule for them.﻿

In the process of fixing it I discovered another small bug relating to this syntax:

fieldname: ${{fake.foo}}

This was causing an exception in the new Native Types mode of Snowfakery. Now it's fixed.


boakley (approved):
I don't see any red flags. This looks pretty straight-forward and I appreciate
the comments in the code related to the jinja quirk.

prescod:
Closes #630

PR #639, opened 2022-04-09
Deep random_references and allow random_references to nicknames

prescod:
Previously the random reference feature was incompatible with two other Snowfakery features.

1. Random References to nicknames did not work:

```
      - object: Parent
         count: 10
        nickname: Bilbo
      - object: Child
        count: 2
        fields:
            A_ref:
              random_reference: Bilbo
```
Now this reference works.

2. Indirection through Random References now works:

```
              - object: Parent
                count: 10
                fields:
                  name:
                    fake: Name

              - object: Child
                fields:
                  pa:
                    random_reference: Parent
                  paps_naame: ${{pa.name}}
```


mitchwd:
Really appreciate your work here @prescod. This feature will make a huge difference for us at Lumary. 🙏 Thank you!

jofsky (changes requested):

jofsky (dismissed):
Approving on condition that default random_references is set to list as discussed and follow up discussion!

docs/index.md
```diff
@@ -594,9 +594,16 @@ The `random_reference` property creates a reference to a random, existing row fr
       random_reference: Owner
 ```
 
-The selected row can be any one that matches the object type and was already created in the current iteration of the recipe. For example, this previous recipe was executed twenty times (iterations) to generate 200 `Pets` and `Owners`. The selected rows in the first iteration would be one of the first ten `Owners`, and the selected rows in the last iteration would be one of the last ten.
+The selected row can be any one that matches the object type (or nickname) and was already created in the current iteration of the recipe. For example, this previous recipe was executed twenty times (iterations) to generate 200 `Pets` and `Owners`. The selected rows in the first iteration would be one of the first ten `Owners`, and the selected rows in the last iteration would be one of the last ten.
```
pconrad-sfdo, line 597: This whole section is confusing. I think what it's saying is it pulls a random row by object, but the object has to be created for it to exist (and get selected).  And if a recipe for 10 pets and owners gets executed 20 times, the pets that get assigned to owners created during the same iteration. But I'm not sure. However, the change that's being made (adding "(or nickname)" seems fine, so it doesn't make sense to block on this.

pconrad-sfdo:
**Here's a quick attempt to make the random_reference section a little clearer:**

The `random_reference property` creates a reference to a random, existing row from a table.

```
- object: Owner
  count: 10
  fields:
    name:
      fake: Name
- object: Pet
  count: 10
  fields:
    ownedBy:
      random_reference: Owner
```

To create a reference, `random_reference` looks for a row created in the current iteration of the recipe and matching the specified object type or nickname. In the above recipe, each `random_reference` specified in `ownedBy` will point to one of the ten `Owner` objects created in the same iteration. If you iterate over the recipe multiple times, in other words, each `Pet` object will be matched with one of the ten `Owner` objects created during the same iteration.

If `random_reference` finds no matches in the current iteration, it looks in previous iterations. This can happen, for example, when you try to create a reference to an object created with the `just_once` flag. Snowfakery cannot currently generate a random_reference to a future iteration of a recipe. 

Performance tip: Tables and nicknames that are referred to by `random_reference` are "indexed", which makes them slightly slower to generate than normal. This should seldom be a problem in practice, but if you experience performance problems you could switch to a normal reference to see if that improves things.

docs/index.md
```diff
@@ -594,9 +594,13 @@ The `random_reference` property creates a reference to a random, existing row fr
       random_reference: Owner
 ```
 
-The selected row can be any one that matches the object type and was already created in the current iteration of the recipe. For example, this previous recipe was executed twenty times (iterations) to generate 200 `Pets` and `Owners`. The selected rows in the first iteration would be one of the first ten `Owners`, and the selected rows in the last iteration would be one of the last ten.
+To create a reference, `random_reference` looks for a row created in the current iteration of the recipe and matching the specified object type or nickname. In the above recipe, each `random_reference` specified in `ownedBy` will point to one of the ten `Owner` objects created in the same iteration. If you iterate over the recipe multiple times, in other words, each `Pet` object will be matched with one of the ten `Owner` objects created during the same iteration.
```
jofsky, line 597:
```suggestion
To create a reference, `random_reference` looks for a row created in the current iteration of the recipe that matches the specified object type or nickname. In the above recipe, each `random_reference` specified in `ownedBy` will point to one of the ten `Owner` objects created in the same iteration. If you iterate over the recipe multiple times, in other words, each `Pet` object will be matched with one of the ten `Owner` objects created during the same iteration.
```

jofsky (approved):

PR #642, opened 2022-04-13
Make Snowfakery compatible with latest CCI

prescod:
Snowfakery's integration testing infrastructure is based on CumulusCI's. CumulusCI's now depends on client apps declaring the `salesforce_serializer`. This PR declares a `salesforce_serializer` which is just the default VCR serializer. Snowfakery doesn't need the compressed serializer yet.


jofsky (approved):
When are we going to implement a compressed serializer?

PR #660, opened 2022-05-11
Dependencies Update

prescod:


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         run: |
           python -VV
           python -m pip install --upgrade pip
-          make dev-install
+          pip install -r requirements_dev.txt
```
boakley, line 27: You removed the use of dev-install but left it in Makefile and it's still referenced in the README and CONTRIBUTING files. Was that all intentional?

boakley (approved):
I left one small comment, but otherwise it looks good to me.

.github/workflows/ci.yml
prescod: I'm hoping in a week or so `make dev-install` will work again when pip-tools is fixed.

jstvz: > I'm hoping in a week or so `make dev-install` will work again when pip-tools is fixed.


I don't think that's necessary. `make dev-install` is a target is for our convenience, and wasn't doing anything special in our CI builds. (The reason it calls `pip-sync` is to synchronize a virtualenv with the requirements files generated by `pip-compile`, but there's no venv to synchronize in the CI runner).

PR #663, opened 2022-05-12
Boolean values should be output as bool in JSON.

prescod:
Boolean values should be output as-such in JSON.


boakley (approved):
looks good to me.

PR #665, opened 2022-05-12
Deps updates and dev release

prescod:
Remove all version pins from Snowfakery to make it simpler to do the same for CCI.


Br4nd0R (approved):
Thanks for doing this @prescod!

PR #676, opened 2022-06-03
Keep track of nicknames for counters

prescod:
This case failed previously:

```
      - object: A
        just_once: True
      - object: A
      - object: A
        nickname: nicky
        fields:
          name: nicky
      - object: B
        count: 2
        fields:
            A_ref:
              random_reference: nicky
            nameref: ${{A_ref.name}}
```

It would not find "nicky" properly. It would find one of the other "A" objects.

### DIFF
--- a/snowfakery/row_history.py
+++ b/snowfakery/row_history.py
@@ -49,6 +49,7 @@ class RowHistory:
 
         if nickname:
             nickname_id = self._get_nickname_id(tablename, nickname)
+            self.table_counters[nickname] = nickname_id
         else:
             nickname_id = None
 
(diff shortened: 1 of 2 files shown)


snowfakery/row_history.py
```diff
@@ -93,13 +94,17 @@ def random_row_reference(self, name: str, scope: str, unique: bool):
                 warnings.warn("Global scope is an experimental feature.")
                 self.already_warned = True
             min_id = 1
+        elif nickname:
```
jofsky, line 97: why elif here?

jofsky (approved):

PR #682, opened 2022-06-06
Pickle Refactor

prescod:
1. Move Snowfakery-specific _DISPATCH_TABLE and _SAFE_CLASSES up a level to the place where other Snowfakery-specific code lives. Otherwise the pickler is Snowfakery-agnostic and does not even import other Snowfakery classes.

2. Move the pickling and unpickling stuff into a class so I can inject the _DISPATCH_TABLE and _SAFE_CLASSES. Also allows me to do away with a Class variable called `RestrictedUnpickler.count` which was ugly.


jkasturi-sf (approved):
Looks good

PR #683, opened 2022-06-07
Start of architectural docs.

prescod:


docs/arch/ArchIndex.md
```diff
@@ -0,0 +1,249 @@
...
+    reference: Account
+```
+
+Imagine if we needed to run this Snowfakery recipe 10,000,000 times in 12 simultaneous processes. Even so, there is one execution of the recipe which is special and that's the= first one, where the Tata Account is created.
+
+So Snowfakery would run it once snapshot the "continuation state" and then fan that continuation state out to the 12 simultaneous processes.
+
+When reading Snowfakery code, you must always think about the lifetime of each data structure:
+
+* Will it survive for a single iteration, like local variables? We call these Transients.
```
jofsky, line 60: Should we define the terms first before introducing how to find them/what to look for to establish a foundation to build on?

jofsky (approved):

docs/arch/ArchIndex.md
prescod: I don't totally follow the suggestion @jofsky

Iteration is defined on line 16.
Continuation is define on line 18.

Did you mean these terms or others?

jofsky: I meant more the more explicit definitions with headers but I understand the desired progression. Maybe we should order the sections in the order of which they were brought up?

PR #700, opened 2022-07-11
Serialize just once random_referenced objects

prescod:
If a just_once template created multiple objects, and was randomly referenced, 
only the most recently created object was saved across continuations. Now they 
are all saved.

### DIFF
--- a/snowfakery/data_generator_runtime.py
+++ b/snowfakery/data_generator_runtime.py
@@ -36,7 +36,7 @@ ParseResult = "snowfakery.parse_recipe_yaml.ParseResult"
 
 
 # save every single object to history. Useful for testing saving of datatypes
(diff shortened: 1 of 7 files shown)


snowfakery/data_generator_runtime.py
```diff
@@ -130,6 +130,7 @@ def __init__(
         # They survive iterations and continuations.
         self.persistent_nicknames = {}
         self.persistent_objects_by_table = {}
+        self.persistent_random_referenceable_objects = []
```
jofsky, line 133: Is there a reason this isn't a dict?

jofsky (dismissed):

snowfakery/data_generator_runtime.py
prescod: Does b671d0f clarify, @jofsky

PR #727, opened 2022-09-15
Fix URL for CumulusCI  Get Started page

BrettMN:

PR #730, opened 2022-09-18
Add diff-cover coverage checking to Snowfakery.

prescod:
Add `diff-cover` coverage checking to Snowfakery.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ env:
   COVERALLS_PARALLEL: true
   COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
   SF_MKDOCS_BUILD_LOCALES: "False"
+  BASE_REF: main
 
 jobs:
   test:
@@ -44,7 +45,7 @@ jobs:
     steps:
       - uses: "actions/checkout@v2"
       - run: |
-          git fetch --no-tags origin $GITHUB_BASE_REF:$GITHUB_BASE_REF
+          git fetch --no-tags origin $BASE_REF:$BASE_REF
 
       - uses: "actions/setup-python@v1"
         with:
@@ -65,9 +66,9 @@ jobs:
         run: |
           coverage xml
           coverage report
-          echo $GITHUB_BASE_REF
-          git diff HEAD..$GITHUB_BASE_REF
-          diff-cover coverage.xml --fail-under 100 --compare-branch=$GITHUB_BASE_REF --diff-range-notation=.. --show-uncovered --markdown-report coverage.md
+          echo $BASE_REF
+          git diff HEAD..$BASE_REF
+          diff-cover coverage.xml --fail-under 100 --compare-branch=$BASE_REF --diff-range-notation=.. --show-uncovered --markdown-report coverage.md
           cat coverage.md >> $GITHUB_STEP_SUMMARY
 
   faker_docs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,11 +62,13 @@ jobs:
       - name: Run Tests
         run: |
           python -m coverage run -m pytest
+
+      - name: Check Coverage
+        run: |
           coverage xml
           coverage report
           git branch -a
           diff-cover coverage.xml --fail-under 100 --compare-branch=origin/main --diff-range-notation=.. --show-uncovered --markdown-report coverage.md
-          echo "== Coverage ==" >> $GITHUB_STEP_SUMMARY
           cat coverage.md >> $GITHUB_STEP_SUMMARY
 
   faker_docs:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -55,7 +60,16 @@ jobs:
...
+        run: |
+          python -m coverage run -m pytest
+
+      - name: Check Coverage
+        run: |
+          coverage xml
+          coverage report
+          git branch -a
+          diff-cover coverage.xml --fail-under 100 --compare-branch=origin/main --diff-range-notation=.. --show-uncovered --markdown-report coverage.md
+          cat coverage.md >> $GITHUB_STEP_SUMMARY
```
jstvz, line 72: This also works:
```suggestion
          diff-cover coverage.xml --fail-under 100 --compare-branch=origin/main --diff-range-notation=.. --show-uncovered --markdown-report $GITHUB_STEP_SUMMARY
```

.github/workflows/ci.yml
```diff
@@ -55,7 +60,16 @@ jobs:
...
       - name: Run Tests
-        run: python -m pytest
+        run: |
+          python -m coverage run -m pytest
+
+      - name: Check Coverage
+        run: |
+          coverage xml
+          coverage report
+          git branch -a
```
jstvz, line 70: For debugging?

.github/workflows/ci.yml
```diff
@@ -43,6 +43,11 @@ jobs:
     runs-on: sfdc-ubuntu-latest
     steps:
       - uses: "actions/checkout@v2"
+      - run: |
+          git fetch --no-tags origin main
+          git switch main
```
jstvz, line 48: This assumes the PR is against main, is it for backwards compatibility?

The `$GITHUB_BASE_REF` env var is available as an alternative. If you use the refspec form (`[...] origin $GITHUB_BASE_REF:$GITHUB_BASE_REF`) you can set up the local tracking branch during the fetch.

jstvz (commented):
One blocking query about defaulting to main as the base branch.

Two other questions came up during the review:
1. We'll still have steps that use `coveralls`, including the final workflow job `coveralls_done`. Did you mean to pull those out?
2. FYI, there's no tox target for `py310` defined, so we aren't running unit tests against 3.10/Linux as currently configured.

.github/workflows/ci.yml
prescod: @jstvz I did it in two steps in case some previous step writes to GITHUB_STEP_SUMMARY at some point. `>>` appends and `-markdown-report $GITHUB_STEP_SUMMARY` overwrites

.github/workflows/ci.yml
prescod: Yes I can remove that now.

prescod:
@jstvz "We'll still have steps that use coveralls, including the final workflow job coveralls_done. Did you mean to pull those out?"

I decided to leave them until I can figure out how to get the "badge" implemented with the new coverage checker.

.github/workflows/ci.yml
prescod: Only works on certain kind of github action events, and "push" is not one of them. I think it's acceptable to just compare against main for now.

.github/workflows/ci.yml
prescod: Gone now.

.github/workflows/ci.yml
jstvz: Each step (i.e., `- run:`) has it's own step summary. All steps are concatenated into the final output.

.github/workflows/ci.yml
jstvz: Right. We switched to `pull_request` in SFDO-Tooling/CumulusCI@aa74f8d because of this. The problem is the noise: PRs against child branches will trigger spurious failures if main is updated with lines that aren't in the coverage report.

Other reasons to switch to `pull_request_target`:
1. External contributors don't push to the repo and [won't trigger workflows](https://github.com/SFDO-Tooling/Snowfakery/pull/727/commits/e8908e4a8af92cda72ba8d97444465354eafa161).
2. Build minutes cost money since the switch to premium runners.

That said, I'm fine with switching to a different event in a separate PR.

.github/workflows/ci.yml
```diff
@@ -55,7 +59,17 @@ jobs:
...
+        run: |
+          python -m coverage run -m pytest
+
+      - name: Check Coverage
+        run: |
+          coverage xml
+          coverage report
+          echo $BASE_REF
+          git diff HEAD..$BASE_REF
+          diff-cover coverage.xml --fail-under 100 --compare-branch=$BASE_REF --diff-range-notation=.. --show-uncovered --markdown-report coverage.md
```
jstvz, line 71: Looks like `--compare-branch` defaults to `origin/main`, so `$BASE_REF` isn't necessary.

jstvz (approved):

PR #737, opened 2022-09-30
Speed up docs build with caching

prescod:
This cache was turned off because of a Faker change and the consequences were bad.

With caching, building the Faker docs took about 8[ minutes](https://github.com/SFDO-Tooling/Snowfakery/actions/runs/1650674753/jobs/2229420189).

With no caching, they took 30 [minutes](https://github.com/SFDO-Tooling/Snowfakery/actions/runs/3140455319).

Now they are back to 8 [minutes](https://github.com/SFDO-Tooling/Snowfakery/actions/runs/3155971564/jobs/5135195733)


prescod (commented):
Test

TheBitShepherd (approved):

PR #740, opened 2022-10-06
Add type checking to Continuous Integration for Snowfakery

prescod:
Made some changes to coveragerc to not cover type-checker-only branches.

While I was in CI I made small changes relating to the benchmarker. It now writes a summary of its output to the github actions output screen.

I updated some dependencies.

I added a Typeguard workaround.

Deleted an old file that was broken and not important.

Cleaned up some type signatures.


snowfakery/tools/storage_bench.py
```diff
@@ -1,69 +1,4 @@
...
-        assert rh.random_row_reference("table", "current-iteration", False).blah
-        assert rh.random_row_reference("Nicknamed", "current-iteration", False).blah
-        assert rh.random_row_reference("Nicknamed2", "current-iteration", False).blah
-        assert rh.random_row_reference("Sparse", "current-iteration", False).blah
-    end = time()
-    print("Loaded", end - start)
-
-
-main(COUNT)
+# Deleted after 2f589a4a44f7ff8e9a4650e978d24839d47f61d9
```
jstvz, line 3: I was going to complain about leaving this file around, but I benchmarked it and it made no difference whatsoever. 🤐

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.pyright]
+include = [
```
jstvz, line 2: This requires some discipline to maintain. Do you anticipate our flavor of gradual typing being a series of focused efforts to annotate a given module and then add it here?

jstvz (approved):
LGTM

---
Did you ever try [PyAnnotate](https://github.com/dropbox/pyannotate), or something like it, to auto-generate type annotations?

snowfakery/tools/storage_bench.py
prescod: Good one dad!

pyproject.toml
prescod: Yes. That's pretty much what I envision. I use wildcards when I can.

I guess in retrospect maybe I could use opt-out instead of opt-in. I'll probably change it to do that later.

PR #778, opened 2022-12-24
Snowfakery's Debug and CSV outputs now use the same format that Salesforce prefers

prescod:
Many, many very simple changes.

1. Snowfakery's "DebugOutput" which people use for debugging their Snowfakery recipes, used a different syntax than Salesforce did. Now it uses the same, which is less confusing. "2024-01-01 12:02:02+00:00" becomes "2024-01-01T12:02:02+00:00" (T in the middle).
2. CSV Output also uses that format now. This [fixes a bug in CumulusCI](https://gus.lightning.force.com/lightning/r/ADM_Sprint__c/a0lEE000000LOTxYAO/view#:~:text=Select%20Row%20Checkbox-,W%2D12170097,-Bug) where CumulusCI uses [the CSV output stream](https://github.com/SFDO-Tooling/CumulusCI/blob/b6238d02982846da4cae9a6b0a581f250c1005f1/cumulusci/tasks/bulkdata/update_data.py#L121) for updates in Salesforce. Because the CSV output stream will now use Salesforce-friendly date-times, it will fix the bug.
3. To avoid rewriting every test that used the old format, I created a new OutputStream called `SimpleFileOutputStream` which is used by tests whereas the `DebugOutput` is for interactive use.


TheBitShepherd (approved):
Looks good!

PR #805, opened 2023-02-03
Use CumulusCI source distribution in CI

jstvz:
We removed test subdirectories from our binary distribution of CumulusCI in SFDO-Tooling/CumulusCI/#3476. This broke the `with_cumulusci` job in our CI workflow because of our dependency on a pytest plugin upstream. This PR forces the use of CumulusCI's source distribution (actually the main branch entirely), which does include the subdirectories we need.

We'll switch to using the source distribution after the next release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           python -VV
           python -m pip install --upgrade pip
           pip install -r requirements_dev.txt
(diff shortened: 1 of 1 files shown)


jstvz:
Fixed upstream in https://github.com/SFDO-Tooling/CumulusCI/pull/3520

prescod (approved):

PR #844, opened 2023-03-29
Improve error message for random_record query into Salesforce org

prescod:
When the user queries an org that has no records, the error message was poor. Now it is clear.

Also: isort. (sort imports)


examples/salesforce_soql_example_no_records.recipe.yml
```diff
@@ -0,0 +1,7 @@
+- plugin: snowfakery.standard_plugins.Salesforce.SalesforceQuery
+- object: Contact
+  fields:
+    FirstName: Suzy
+    LastName: Salesforce
+    AccountId:
+      SalesforceQuery.random_record: Contract
```
prescod, line 7: Change "Contract" to Account

PR #855, opened 2023-04-12
Update deps

prescod:
Update dependencies and remove TypeGuard.

We can use Pyright now as a simpler alternative to TypeGuard, and TypeGuard was causing really weird problems.


jkasturi-sf (approved):
Looks good

PR #881, opened 2023-05-04
SQLAlchemy 2 compatible code

prescod:
In SQL Alchemy 2, 

 1. Metadata objects are not bound to engines. You don't bind until you create tables or reflect schema back into the Metadata object.
 2. One must make transactions more explicitly to ensure that records are saved
 3.  Rows cannot as easily be converted to dicts. You must use the ._mapping property rist
 4. You cannot execute raw strings: you must wrap them in text() objects.
 5. select statements do not take lists of tables. They take a single table or other similar object.


jkasturi-sf (approved):
Looks good

PR #883, opened 2023-05-07
Make macro-nested fields available to the top level

prescod:
Fix issue from #205.

Fields in a macro in a macro are not available to the formulas of the top-level.


jkasturi-sf (approved):
Looks good

PR #886, opened 2023-05-08
Add Snowfakery 3 header to all examples

prescod:
We want people to start making snowfakery_version: 3 recipes
so we can get rid of snowfakery_version: 2. Having examples
and documents in that format is one step towards that.


jkasturi-sf (approved):
Looks good

PR #1029, opened 2024-04-05
Clean up temporary directories after closing iterator

### LINKED ISSUE #901: Windows 237 NotADirectory Error with Salesforce Query
I'm running Windows with VSCode-based Powershell and snowfakery 3.5.
When I try to run a [snowfakery recipe](https://github.com/SFDO-Community-Sprints/Snowfakery-Recipe-Templates/blob/main/snowfakery_samples/OSC/MembershipSchemaAndBenefits_PricebookProducts.recipe.yml) that queries Salesforce records, I receive an error message.

This appears to be the crux of the issue: 
`NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\ALLISO~1\\AppData\\Local\\Temp\\tmp9b9uku52\\queryresults.db'`

Full documentation:
https://gist.github.com/allisonletts/994828955eb9c266936f73b4d9e52b4b

prescod:
Fixes #901

What was happening on Windows is that that the temporary directory was being garbage collected before the open file handle temporary file inside of it. This was probably also handling on *ix, but it was harmless.

On Windows, the temporary directory object needs to live a bit longer than the local tempfile, which needs to live a bit longer than than the SQL Alchemy connection to the tempfile.

So I use an ExitStack to inject cleanup code which is run before the connection is closed and the file handle is released.

All of this is triggered by the garbage collector letting go of the iterator when the recipe is finishing up.

It might be better to do a deterministic, explicit, top-down cleanup, but that should probably be a different PR.

### DIFF
--- a/snowfakery/standard_plugins/Salesforce.py
+++ b/snowfakery/standard_plugins/Salesforce.py
@@ -330,26 +330,14 @@ class SOQLDatasetImpl(DatasetBase):
                 f"Unable to query records for {query}: {','.join(qs.job_result.job_errors)}"
             )
 
-        self.tempdir, self.iterator = create_tempfile_sql_db_iterator(
+        tempdir, iterator = create_tempfile_sql_db_iterator(
             iteration_mode, fieldnames, qs.get_results()
         )
-        return self.iterator
+        iterator.cleanup.push(tempdir)
+        return iterator
 
     def close(self):
-        if self.iterator:
-            self.iterator.close()
-            self.iterator = None
-
-        if self.tempdir:
-            self.tempdir.cleanup()
-            self.tempdir = None
-
-    def __del__(self):
-        # in case close was not called
-        # properly, try to do an orderly
-        # cleanup
-        self.close()
-
+        pass
 
 def create_tempfile_sql_db_iterator(mode, fieldnames, results):
     tempdir, db_url = _create_db(fieldnames, results)
--- a/snowfakery/standard_plugins/datasets.py
+++ b/snowfakery/standard_plugins/datasets.py
@@ -66,8 +66,16 @@ class DatasetIteratorBase(PluginResultIterator):
     Subclasses should implement 'self.restart' which puts an iterator into 'self.results'
     """
 
+    def __init__(self, repeat):
+        # subclasses can register stuff to be cleaned up here.
+        self.cleanup = ExitStack()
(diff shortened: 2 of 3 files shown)


snowfakery/standard_plugins/Salesforce.py
```diff
@@ -330,26 +330,14 @@ def _load_dataset(self, iteration_mode, rootpath, kwargs):
                 f"Unable to query records for {query}: {','.join(qs.job_result.job_errors)}"
             )
 
-        self.tempdir, self.iterator = create_tempfile_sql_db_iterator(
```
prescod, line 333: The Plugin object is a singleton.

It can create an arbitrary number of query iterators. It should never have saved them on "self."

But the temporary directory object needs to live a bit longer than the local tempfile, which needs to live a bit longer than than the SQL Alchemy connection to the tempfile.

So I use an ExitStack to inject cleanup code which is run before the connection is closed and the file handle is released.

All of this is triggered by the garbage collector letting go of the iterator when the recipe is finishing up.

It might be better to of a deterministic, explicit, top-down cleanup, but that should probably be a different PR.

snowfakery/standard_plugins/Salesforce.py
```diff
@@ -330,26 +330,14 @@ def _load_dataset(self, iteration_mode, rootpath, kwargs):
...
+        tempdir, iterator = create_tempfile_sql_db_iterator(
             iteration_mode, fieldnames, qs.get_results()
         )
-        return self.iterator
```
prescod, line 336: Here I am adding to the ExitStack

snowfakery/standard_plugins/Salesforce.py
```diff
@@ -330,26 +330,14 @@ def _load_dataset(self, iteration_mode, rootpath, kwargs):
...
+        iterator.cleanup.push(tempdir)
```
prescod, line 336: The temporary directory object needs to live a bit longer than the local tempfile, which needs to live a bit longer than than the SQL Alchemy connection to the tempfile.

So I use an ExitStack to inject cleanup code which is run before the connection is closed and the file handle is released.

This line adds to the ExitStack.

All of this is triggered by the garbage collector letting go of the iterator when the recipe is finishing up.

It might be better to of a deterministic, explicit, top-down cleanup, but that should probably be a different PR.

PR #1030, opened 2024-04-06
Updated to Pydantic 2 (superset of Jason's)

prescod:
Obsoletes #1018


prescod:
@jstvz 

The current test failure is because CCI is on Pydantic 1 while Snowfakery is on Pydantic 2.

Do you want to:

a) upgrade CCI and Snowfakery to Pydantic 2 in a single coordinated release or

b) support Pydantic 1 and 2 simultaneously in Snowfakery so that they can upgrade in a staggered manner?

prescod:
Obviously b) involves ugly cruft that will need to be removed eventually.

jstvz:
> @jstvz
> 
> The current test failure is because CCI is on Pydantic 1 while Snowfakery is on Pydantic 2.
> 
> Do you want to:
> 
> a) upgrade CCI and Snowfakery to Pydantic 2 in a single coordinated release or
> 
> b) support Pydantic 1 and 2 simultaneously in Snowfakery so that they can upgrade in a staggered manner?

Option (a) is my preference.

prescod:
@jstvz . Okay, ping me when the time comes if there are challenges on the Snowfakery side. It actually doesn’t thave a lot of Pydantic stuff in it and arguably what it does have should move to the CCI repo because it relates to CCI.

kwendel:
Is there any progress on this PR? We are using snowfakery in our project and would love to update it to pydantic V2 like our other dependencies.

I would be willing to help if that is possible

jstvz:
@kwendel Thanks for the reminder. This PR is currently blocked by a dependency on CumulusCI, which also needs a Pydantic V2 upgrade. I'm bumping the priority with the team internally and will update you here.

PR #1031, opened 2024-04-06
Let mkdocs float

prescod:
Switch to the latest mkdocs and let the version float in the future.


mkdocs.yml
```diff
@@ -1,8 +1,9 @@
 site_name: Snowfakery documentation
 site_url: ""
 use_directory_urls: False
-theme: readthedocs
-custom_dir: custom_theme
+theme:
+  name: readthedocs
+  custom_dir: custom_theme
```
jstvz, line 6: `ERROR   -  Config value 'theme': The path set in custom_dir ('custom_theme') does not exist.`

```suggestion
```

jstvz (changes requested):

jstvz (approved):

PR #1032, opened 2024-04-06
Let SQLAlchemy float

prescod:
Let SQL Alchemy version float to 2.0.

I've smoke tested against 2.0

Mostly the changes are about resource management. Lots of introduction of context managers.

Also a few assertions to make type checker happy.


prescod:
@jstvz I think this is ready for review now. The tests are probably fixed.

jstvz (approved):

PR #1093, opened 2024-11-15
Update ReadTheDocs configuration

jstvz:
- update to Ubuntu 22.04 and Python 3.12,
- add uv setup and mkdocs build commands.


mgrandhi (approved):
LGTM

